PR #7, opened 2025-04-09
Fix/minor fixes

griffin-norris:
- remove some debug prints
- remove a (commented out) circular import
- add more type hinting for `Config` type
- rename test to `test_examples`


griffin-norris:
lgtm

PR #9, opened 2025-04-09
Feature/better constraint decorators

griffin-norris:
- allow `nodal` constraint decorator to include which nodes to apply for, defaults to all nodes
- slight change to how CTCS penalty function is applied
- Addition of Huber and smooth ReLU CTCS penalty functions


griffin-norris:
merging despite CI failing, issue seems to be due to previous merges in main, will fix in future PR

Issue #11, opened 2025-04-10
Custom integrator is super slow


griffin-norris:
Using the jax-based custom integrator results in extremely long initialization time and high ram usage. Causes [`dr_vp`](https://github.com/haynec/OpenSCvx/blob/main/examples/params/dr_vp.py) to not execute on a laptop with 16GB of RAM

@haynec pointed out that this is potentially related to how JAX interacts with lambda functions: https://github.com/jax-ml/jax/discussions/10868

griffin-norris:
Tracked down the slowdown to d1f20771

haynec:
This should be fixed in https://github.com/haynec/OpenSCvx/commit/48d7663b163eea318ac63bc934269ca210398b5f

PR #17, opened 2025-04-12
using low level jax.lax in custom integrator

haynec:


haynec:
@griffin-norris This should address the concerns raised in https://github.com/haynec/OpenSCvx/issues/11

PR #19, opened 2025-04-13
Refactor/ptr post

haynec:
Changelog:
- Removed viewplanning-specific scripts from PTR post and put them into plotting.py


haynec:
@griffin-norris Please take a look and merge if you are good with it.

PR #22, opened 2025-04-13
Refactor/centralized compiling

haynec:
Changelog:
- Moved AOT compiling from discretization.py into trajoptproblem.py


griffin-norris (approved):
Lgtm, in the future we could maybe make the compiling more elegant so that you don't have to remember which subcomponent needs to be assigned the compiled version of the function

Issue #26, opened 2025-04-14
Add Diffrax integrator option


griffin-norris:
Include [Diffrax](https://docs.kidger.site/diffrax/) as an option for integration, compare performance with scipy and existing custom integrator

haynec:
Performs significantly better then scipy, so replace with Diffrax

PR #32, opened 2025-04-15
Feature/diffrax

haynec:
Changelog:
- Added diffrax to packages
- Added diffrax solver and arguments as `problem.params.sim.diffrax_solver` and `problem.params.sim.diffrax_args`

Fixes https://github.com/haynec/OpenSCvx/issues/26

### DIFF
--- a/openscvx/config.py
+++ b/openscvx/config.py
@@ -132,6 +132,8 @@ class SimConfig:
     inv_S_u: np.ndarray = None
     c_u: np.ndarray = None
     diffrax: bool = False
+    diffrax_solver: str = 'Tsit5'
+    diffrax_args: Dict = field(default_factory=dict)
 
     def __post_init__(self):
         self.n_states = len(self.max_state)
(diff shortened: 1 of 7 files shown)


openscvx/config.py
```diff
@@ -132,6 +132,8 @@ class SimConfig:
     inv_S_u: np.ndarray = None
     c_u: np.ndarray = None
     diffrax: bool = False
+    diffrax_solver: str = 'Tsit5'
```
griffin-norris, line 135: does this overlap with the scipy integrator solver choice? Or is there an equivalent? Maybe we could unify the integrator options with some minor parsing depending on which integrator is chosen

griffin-norris (approved):
lgtm with minor comment above that could be addressed in future (or now)

openscvx/config.py
haynec: Good point. I figure keep them separate for now since there is not enough overlap of solvers. I will open an issue for removing scipy all together here, https://github.com/haynec/OpenSCvx/issues/34

PR #80, opened 2025-05-06
Feature/augmented states

haynec:
Changelog:
- Added ability to have multiple augmented states for constraint violation. 
- The `ctcs` decorator now has the additional arguments `nodes = "all", exclusive = False`. Setting nodes to `"all"` will apply the `ctcs` constraint over the full interval. Unless `exclusive` is set to be `True`, the constraints will be grouped into the same augmented state corresponding to their associated nodes. If `exclusive` is `True`, then the constraint will gets its own _exclusive_ augmented state.  To set it over a subinterval set to 'nodes=(1, 4)' where the `tuple` correspond to the beginning node and final node. Note that is the beginning node is 0 it will be fixed to 1 as the LICQ constraint is formulated as from `y_k - y_k-1 <= eps_licq`, see the implementation below.

```python
if nodes[0] == 0:
    start_indx = 1
else:
    start_indx = nodes[0]
constr += [cp.abs(x_nonscaled[i][idx] - x_nonscaled[i-1][idx]) <= params.sim.max_state[idx] for i in range(start_indx, nodes[1])]
```

### DIFF
--- a/examples/params/cinema_vp.py
+++ b/examples/params/cinema_vp.py
@@ -10,7 +10,6 @@ from openscvx.constraints.decorators import ctcs
 n = 12  # Number of Nodes
 total_time = 40.0  # Total time for the simulation
 
-
 fuel_inds = 13  # Fuel Index in State
 t_inds = 14
 s_inds = 6  # Time dilation index in Control
--- a/examples/params/cinema_vp.py
+++ b/examples/params/cinema_vp.py
@@ -136,7 +136,7 @@ problem = TrajOptProblem(
     x_min=min_state,
     u_max=max_control,  # Upper Bound on the controls
     u_min=min_control,  # Lower Bound on the controls
-    ctcs_augmentation_max=1e-8,
+    licq_max=1e-8,
 )
 
 problem.params.prp.dt = 0.1
--- a/openscvx/constraints/decorators.py
+++ b/openscvx/constraints/decorators.py
@@ -2,9 +2,11 @@ from jax import jit, vmap, jacfwd
 import jax.numpy as jnp
 
 
-def ctcs(func: callable, penalty="squared_relu") -> callable:
+def ctcs(func: callable, penalty="squared_relu", nodes = "all", exclusive = False) -> callable:
     """Decorator to mark a function as a 'ctcs' constraint."""
     func.constraint_type = "ctcs"
+    func.nodes = nodes
+    func.exclusive = exclusive
     if penalty == "squared_relu":
         func.penalty = lambda x: jnp.maximum(0, x) ** 2
     elif penalty == "huber":
--- a/openscvx/constraints/decorators.py
+++ b/openscvx/constraints/decorators.py
@@ -2,11 +2,13 @@ from jax import jit, vmap, jacfwd
 import jax.numpy as jnp
 
 
-def ctcs(func: callable, penalty="squared_relu", nodes = "all", exclusive = False) -> callable:
+def ctcs(func: callable, penalty="squared_relu", nodes = "all", exclusive = False, licq_max=None) -> callable:
     """Decorator to mark a function as a 'ctcs' constraint."""
     func.constraint_type = "ctcs"
     func.nodes = nodes
     func.exclusive = exclusive
+    func.licq_max = licq_max
+    
     if penalty == "squared_relu":
         func.penalty = lambda x: jnp.maximum(0, x) ** 2
     elif penalty == "huber":
--- a/openscvx/constraints/decorators.py
+++ b/openscvx/constraints/decorators.py
@@ -2,12 +2,11 @@ from jax import jit, vmap, jacfwd
 import jax.numpy as jnp
 
 
-def ctcs(func: callable, penalty="squared_relu", nodes = "all", exclusive = False, licq_max=None) -> callable:
+def ctcs(func: callable, penalty="squared_relu", nodes = "all", exclusive = False) -> callable:
     """Decorator to mark a function as a 'ctcs' constraint."""
     func.constraint_type = "ctcs"
     func.nodes = nodes
     func.exclusive = exclusive
-    func.licq_max = licq_max
     
     if penalty == "squared_relu":
         func.penalty = lambda x: jnp.maximum(0, x) ** 2
--- a/openscvx/dynamics.py
+++ b/openscvx/dynamics.py
@@ -2,12 +2,19 @@ import jax
 import jax.numpy as jnp
 
 
-def get_augmented_dynamics(dynamics: callable, g_func: callable):
+def get_augmented_dynamics(dynamics: callable, g_func: dict[str, callable]):
     def dynamics_augmented(x: jnp.array, u: jnp.array) -> jnp.array:
         # TODO: (norrisg) only pass user-defined portion of x to dynamics in case user has `-1` or similar indexing in function
         x_dot = dynamics(x, u)
-        y_dot = g_func(x, u)
-        return jnp.hstack([x_dot, y_dot])
+        # y_dot = g_func(x, u)
+        # return jnp.hstack([x_dot, y_dot])
+
+        # Iterate through the g_func dictionary and stack the output each function
+        # to x_dot
+        for key, g in g_func.items():
+            x_dot = jnp.hstack([x_dot, g(x, u)])
+        
+        return x_dot
 
     return dynamics_augmented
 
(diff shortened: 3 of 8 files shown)


openscvx/trajoptproblem.py
```diff
@@ -132,22 +186,23 @@ def __init__(
...
-                )
-            elif constraint.constraint_type == "nodal":
-                sim.constraints_nodal.append(constraint)
-            else:
-                raise ValueError(
-                    f"Unknown constraint type: {constraint.constraint_type}, All constraints must be decorated with @ctcs or @nodal"
-                )
+        sim.constraints_ctcs = constraints_ctcs
+        sim.constraints_nodal = constraints_nodal
+        sim.num_augmented_states = num_augmented_states
```
griffin-norris, line 191: Is it necessary to save the num of augmented states? Couldn't this information be retrieved from the length of the constraint violation slice? As it is now it would be possible to have the `num_augmented_states` and `idx_constraint_violation` disagree which could be a source of bugs

tests/test_examples.py
```diff
@@ -91,8 +91,9 @@ def test_example_problem(name, conf):
     # assertions
     assert sol_cost < conf["max_cost"], "Problem failed with solution cost"
     assert prop_cost < conf["max_cost"], "Problem failed with propagated cost"
-    assert sol_constr_vio  < conf["max_vio"], "Problem failed with solution constraint violation"
-    assert prop_constr_vio < conf["max_vio"], "Problem failed with propagated constraint violation"
+    if name != "dr_vp_nodal":
```
griffin-norris, line 94: Don't check if the name is `dr_vp_nodal`, change it so that the limits for the nodal case are all set to -1, if a limit is -1 then don't use the assert statement.

Directly checking against the name requires us to update this code if the name ever changes

griffin-norris (changes requested):
Minor comments. Mostly curious if the additional args for the ctcs decorator are really necessary like that, and if the if-else block can't be simplified or at least compartmentalized in some way.

openscvx/trajoptproblem.py
```diff
@@ -51,18 +51,72 @@ def __init__(
...
+                raise ValueError(
+                    f"Unknown constraint type: {constraint.constraint_type}, All constraints must be decorated with @ctcs or @nodal"
+                )
+        
+        # Constraints should be added to the list if there are multiple constraints for the same set of nodes
+        # unless they are specified as exclusive
+        constraints_dict = {}
+        shared_ids = {}  # Dictionary to store shared IDs for non-exclusive constraints
+
+        for constraint in constraints_ctcs:
```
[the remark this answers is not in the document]
haynec: This much has already been there from before, 

```python
for constraint in constraints:
            if constraint.constraint_type == "ctcs":
                constraints_ctcs.append((
                    lambda x, u, func=constraint: jnp.sum(func.penalty(func(x[idx_x_true], u[idx_u_true]))), constraint))
            elif constraint.constraint_type == "nodal":
                constraints_nodal.append(constraint)
            else:
                raise ValueError(
                    f"Unknown constraint type: {constraint.constraint_type}, All constraints must be decorated with @ctcs or @nodal"
                )
```

I just moved it around. 

If you have a way of more elegantly dealing with that information encoding this, I am all ears.

openscvx/trajoptproblem.py
haynec: Agreed. I can work on grabbing the number of augmented states from the index

openscvx/trajoptproblem.py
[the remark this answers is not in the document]
griffin-norris: Working on it. After discussion earlier will implement under the hood dataclass. User still interacts throuigh lightweight decorator so they don't have to touch any of the class stuff @haynec

tests/test_examples.py
griffin-norris: I implemented a check there that the assertion is only checked if the conf value is positive. That way when you write the test config you can just specify a negative value for the limit and it will skip that test. So for the nodal case I set `max_vio=-1.0`

fixed in d5a9948

openscvx/trajoptproblem.py
griffin-norris: Removed in [f23e615](https://github.com/haynec/OpenSCvx/pull/80/commits/f23e615f210974a4e7bebff099c15f7eab6b1310)

openscvx/trajoptproblem.py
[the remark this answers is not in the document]
griffin-norris: Fixed in #82

openscvx/constraints/decorators.py
```diff
@@ -1,41 +1,98 @@
...
+@dataclass
+class CTCSConstraint:
+    func: Callable[[jnp.ndarray, jnp.ndarray], jnp.ndarray]
+    penalty: Callable[[jnp.ndarray], jnp.ndarray]
+    nodes: Optional[Sequence[Tuple[int, int]]] = None
+    idx: Optional[int] = None
+
+    def __call__(self, x, u, node):
+        # slice x,u at the true-state indices upstream
+        return jnp.sum(self.penalty(self.func(x, u, node)))
```
griffin-norris, line 18: We can make it so the `__call__` takes `node` as an arg and checks if `node` is inside the node range specified by `self.nodes` -> Then user does not need to define the constraint function with `node` as an arg, removing the burden from the user

openscvx/dynamics.py
```diff
@@ -2,12 +2,18 @@
...
+) -> callable:
+    def dynamics_augmented(x: jnp.array, u: jnp.array, node: jnp.ndarray) -> jnp.array:
+        x_dot = dynamics(x[idx_x_true], u[idx_u_true], node)
+
+        # Iterate through the g_func dictionary and stack the output each function
+        # to x_dot
+        for g in g_funcs:
+            x_dot = jnp.hstack([x_dot, g(x[idx_x_true], u[idx_u_true], node)])
+
+        return x_dot
```
griffin-norris, line 16: Following comment above, can then not pass `node` into dynamics, but should pass `node` into `g`, since it is a `CTCSConstraint` and the `__call__` method expects the node

openscvx/trajoptproblem.py
```diff
@@ -51,18 +53,38 @@ def __init__(
...
+        constraints_nodal = []
+        for constraint in constraints:
+            if type(constraint).__name__ == CTCSConstraint.__name__:
+                constraints_ctcs.append(
+                    constraint
+                )
+            elif type(constraint).__name__ == NodalConstraint.__name__:
+                constraints_nodal.append(
+                    constraint
+                )
```
griffin-norris, line 73: Is this way to check if the type matches necessary once the package is uploaded to / installed from pypi? If not we should make a note to convert it back

examples/params/cinema_vp.py
```diff
@@ -98,11 +97,11 @@ def g_max(x):
...
-    ctcs(lambda x, u: x - max_state),
-    ctcs(lambda x, u: min_state - x),
-    ctcs(lambda x, u: g_min(x)),
-    ctcs(lambda x, u: g_max(x)),
+    ctcs(lambda x, u, node: np.sqrt(2e1) * g_vp(x)),
+    ctcs(lambda x, u, node: x - max_state),
+    ctcs(lambda x, u, node: min_state - x),
+    ctcs(lambda x, u, node: g_min(x)),
+    ctcs(lambda x, u, node: g_max(x)),
 ]
```
griffin-norris, line 105: if above comments are addressed then wouldn't need to specify node as an arg here

griffin-norris: (or in any of these functions

griffin-norris (changes requested):
As discussed offline we should try to make it so the user doesn't need to define `node` as an arg, but to instead automagically add the arg under the hood. I took a stab at implementing this but ran into some errors.

If this is implemented then some of errors in the testing will be resolved automatically.

openscvx/trajoptproblem.py
haynec: Yeah we can leave a note to change it back to `type(constraint) == CTCSConstraint` and `type(constraint) == NodalConstraint`

examples/params/cinema_vp.py
griffin-norris: Resolved by #86

openscvx/dynamics.py
griffin-norris: Resolved by #86

openscvx/constraints/decorators.py
griffin-norris: Resolved by #86

griffin-norris (approved):
LGTM

PR #110, opened 2025-05-11
added custom penalty definition

haynec:
Changelog:
- Re-added custom penalty functions


haynec:
@griffin-norris not sure why this had been removed but I added back in custom penalty functions

PR #151, opened 2025-05-22
Feature/export compiled code

haynec:
Changelog:
- Switched from `jax.jit.().lower().compile()` to `jax.export.export(jax.jit()(*args))` for compiling code as this can be serialized and saved off for use at a later time
- Unique hashes are generated based on the dynamics, propagated dynamics, and constraints for each problem to ensure there is no conflict in the compiled solvers

### DIFF
--- a/openscvx/integrators.py
+++ b/openscvx/integrators.py
@@ -1,8 +1,51 @@
+import os
+os.environ["EQX_ON_ERROR"] = "nan"
+
+
+from typing import Callable, Any, Optional
+
+
+import jax
+import jax.numpy as jnp
+import diffrax as dfx
 from typing import Callable, Any
 
 import jax
 import jax.numpy as jnp
 import diffrax as dfx
+
+from diffrax._global_interpolation import DenseInterpolation
+from jax import tree_util
+
+# Safely check if DenseInterpolation is already registered
+try:
+    # Attempt to flatten a dummy DenseInterpolation instance
+    # Provide dummy arguments to create a valid instance
+    dummy_instance = DenseInterpolation(
+        ts=jnp.array([]),
+        ts_size=0,
+        infos=None,
+        interpolation_cls=None,
+        direction=None,
+        t0_if_trivial=0.0,
+        y0_if_trivial=jnp.array([]),
+    )
+    tree_util.tree_flatten(dummy_instance)
+except ValueError:
+    # Register DenseInterpolation as a PyTree node if not already registered
+    def dense_interpolation_flatten(obj):
+        # Flatten the internal data of DenseInterpolation
+        return (obj._data,), None
+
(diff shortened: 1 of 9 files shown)


openscvx/integrators.py
```diff
@@ -178,32 +222,18 @@ def solve_ivp_diffrax_prop(
...
-        atol (float, optional): Absolute tolerance. Defaults to 1e-6.
-        extra_kwargs (dict, optional): Additional kwargs for `diffeqsolve`.
-
-    Returns:
-        diffrax.Solution: Full solver output, including `ys` and interpolation functions.
-
-    Raises:
-        ValueError: If `solver_name` is not recognized.
+        ...
+        save_times (jnp.ndarray, optional): Time points to evaluate solution at.
```
griffin-norris, line 233: why remove the documentation here? Just to make it slightly cleaner? I still would prefer to have docs for each function so users don't have to read 2 docs to understand 1 funtion

openscvx/trajoptproblem.py
```diff
@@ -31,6 +36,25 @@
...
+                hasher.update(str(func.__code__.co_varnames).encode())
+                hasher.update(func.__name__.encode())
+            else:
+                # Fallback to inspect (less stable)
+                src = inspect.getsource(func)
+                hasher.update(src.encode())
+        except Exception as e:
+            raise ValueError(f"Could not hash function {func}: {e}")
+
+    return hasher.hexdigest()
```
griffin-norris, line 57: Can we move this to a different file?

openscvx/ptr.py
```diff
@@ -126,7 +126,7 @@ def PTR_subproblem(cpg_solve, x_bar, u_bar, aug_dy, prob, params: Config):
     prob.param_dict['u_bar'].value = u_bar
     
     t0 = time.time()
-    A_bar, B_bar, C_bar, z_bar, V_multi_shoot = aug_dy(x_bar, u_bar.astype(float))
+    A_bar, B_bar, C_bar, z_bar, V_multi_shoot = aug_dy.call(x_bar, u_bar.astype(float))
```
griffin-norris, line 130: why do we need `.call` here?

griffin-norris (approved):
LGTM assuming it works, a few minor comments but nothing critical

openscvx/ptr.py
haynec: Because it is a `jax._src.export._export.Exported` instead of `jax.stages.compiled` object.

openscvx/integrators.py
haynec: Sorry, it got lost in the sauce, will readd.

openscvx/trajoptproblem.py
haynec: Done

PR #225, opened 2025-11-13
Fix/propagation dynamics

### LINKED ISSUE #223: Propagation Dynamics No Longer Used Properly

griffin-norris:
- Re-enable user to specify extra propagation dynamics. _Now using the symbolic layer!_
- User now only needs to specify the _extra_ states and their dynamics
- Fixes #223


haynec:
Looks like a good interface.

PR #268, opened 2025-12-03
Fix/integer float value init issues

### LINKED ISSUE #196: Non-Explicit Float Arrays Cause Numerical Issues
- Users currently must write `[0.0, 1.0]` to force the arrays to be floats rather than just `[0, 1]` when instantiating some boundary conditions, initial guesses, _etc._
- Otherwise this can sometimes cause numerical issues
- Should just always convert arrays to floats instead of requiring the user to debug

griffin-norris:
- Fixes #196
- Force boundary conditions, initial guesses, _etc._ of variables to be float datatype
- Force constants, initial values of parameters to be float datatypes
- Force parameter updates to be float datatype

### DIFF
--- a/openscvx/problem.py
+++ b/openscvx/problem.py
@@ -80,8 +80,9 @@ class _ParameterDict(dict):
         # Sync to internal dict for JAX
         self._internal_dict[key] = value
         # Sync to CVXPy if it exists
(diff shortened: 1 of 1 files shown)


haynec (approved):
All good to me

PR #273, opened 2025-12-04
added structural figure

haynec:


griffin-norris (approved):
LGTM. Still think this figure is slightly cluttered but is overall a good addition

Issue #263, opened 2025-12-01
3 DoF PDG Example fails to Converge Sometimes


griffin-norris:
- https://github.com/haynec/OpenSCvx/actions/runs/19791459357

griffin-norris:
fixed in `d04536b1b95eb4f1e6c9f43acc99001cc6f049cd`

PR #277, opened 2025-12-04
Fix/zoh control propagation

haynec:
Changelog:
- Fixed issue where ZOH control would not propagate correctly
- Fixes https://github.com/haynec/OpenSCvx/issues/263


griffin-norris (approved):
LGTM

PR #286, opened 2025-12-07
Feature/problem idempotency improvements

### LINKED ISSUE #123: Add Documentation For TrajOptProblem

### LINKED ISSUE #257: Add problem.reset
We should add a `problem.reset` method which resets the guesses for all states and control (including the internal constraint integrator states) and weights.

griffin-norris:
- Fixes #257
- Makes `Problem` idempotent by:
  - Consolidating scattered SCP iteration state (`scp_k`, `scp_J_tr`, `scp_J_vb`, `scp_J_vc`, `scp_trajs`, `scp_controls`) into a unified `SolverState` object
  - Preserve idempotency by creating separate compiled instances (`_compiled_dynamics`, `_compiled_constraints`) while keeping original un-vmapped/un-JIT'd versions in `_lowered`
  - Separate active solver state (`_state`) from converged solution (`_solution`) for clearer semantics and post-processing workflow
  - This means that now it is safe to call
```python
problem.initialize()
problem.initialize() # No error, produces same results
results = problem.solve()
results = problem.solve() # No error, produces same results
```
- Add `Problem.reset()` method to enable re-running optimization from initial conditions without full re-initialization
- Breaking: `Problem.post_process()` has also been changed to be idempotent
  - Now uses the internal `Problem._solution` instead of taking a `results: OptimizationResults` arg
  - This is a breaking change since the signature is now:
```python
results = problem.solve() # Partial results
results = problem.post_process() # Full results
``` 
- Refactor profiling logic into reusable profiling_start() and profiling_end() utilities
- Fixes #123 -> adds more docstrings to `Problem`
- Improve encapsulation by making OCP and solvers private (`_optimal_control_problem`, `_discretization_solver`, etc.)
- Pass parameters to solver constructors for better compilation caching

### DIFF
--- a/openscvx/solver_state.py
+++ b/openscvx/solver_state.py
@@ -103,7 +103,7 @@ class SolverState:
         V_final = V[:, -1].reshape(-1, i4)
 
         # Extract propagated state (first n_x elements of each row)
-        return V_final[:, :self.n_x]
+        return V_final[:, : self.n_x]
 
     @property
     def A_d(self) -> np.ndarray:


openscvx/solver_state.py
haynec: We should track additional items in here, specifically storing the discretization results (`A_d, B_d, C_d, x_prop`) and copies of the `cvxpy.problem.var_dict` per iteration. This paves the way for autotuning nicely.

haynec (approved):
LGTM subject to my comments. I added in a reset problem button to the realtime examples and verified it works as expected.

PR #351, opened 2026-01-27
Feature/augmented lagrangian

haynec:
Changelog:
- Added the Augmented Lagrangian method for autotuning the proximal and linear penalty weights.
- Updated examples fixes https://github.com/OpenSCvx/OpenSCvx/issues/287
- Adds https://github.com/OpenSCvx/OpenSCvx/issues/278

---

> [!NOTE]
> **High Risk**
> High risk because it changes core SCP/PTR iteration behavior (accept/reject logic, weight updates, discretization/state handling) and renames a widely-used tuning parameter, which can affect solver convergence and downstream user code despite added tests and docs updates.
> 
> **Overview**
> Adds a new SCP autotuning API (`AutotuningBase`) with strategies like `AugmentedLagrangian`, `ConstantProximalWeight`, and `RampProximalWeight`, and wires it into `PenalizedTrustRegion` so weight/cost updates happen via `settings.scp.autotuner`.
> 
> Renames the trust-region weight from `w_tr` to `lam_prox` across config, lowering, solver parameters, realtime GUIs, docs, and examples; cost relaxation is now configured on the autotuner (e.g., `lam_cost_drop`/`lam_cost_relax`).
> 
> Refactors solver iteration state: introduces `CandidateIterate`, stores unpacked discretizations via `DiscretizationResult`, tracks new convergence histories (reductions, acceptance ratio, `lam_prox`), improves constraint linearization shape handling, and adds new plotting (`plot_scp_convergence_histories`) plus adaptive CLI printing.
> 
> Adds a new realtime drone racing base problem/example, pins `cvxpy` to `1.7.5` temporarily until https://github.com/cvxpy/cvxpy/issues/3092 is addressed in the 1.8.0 release, and introduces extensive unit tests for autotuning.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 786d0f764b358249a28319cbcadb76bfad666cf7. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/openscvx/algorithms/autotuning.py
+++ b/openscvx/algorithms/autotuning.py
@@ -208,14 +208,26 @@ class AugmentedLagrangian(AutotuningBase):
         Args:
             rho_init: Initial penalty parameter for constraints. Defaults to 1.0.
             rho_max: Maximum penalty parameter. Defaults to 1e6.
-            rho_increase: Factor to increase penalty when constraints violated.
+            gamma_1: Factor to increase trust region weight when ratio is low.
                 Defaults to 2.0.
-            rho_decrease: Factor to decrease penalty when constraints satisfied.
+            gamma_2: Factor to decrease trust region weight when ratio is high.
                 Defaults to 0.5.
-            mu_init: Initial penalty parameter for virtual control. Defaults to 1.0.
-            mu_max: Maximum penalty parameter for virtual control. Defaults to 1e6.
-            mu_increase: Factor to increase mu when dynamics violated. Defaults to 2.0.
-            mu_decrease: Factor to decrease mu when dynamics satisfied. Defaults to 0.5.
+            eta_0: Acceptance ratio threshold below which solution is rejected.
+                Defaults to 1e-2.
+            eta_1: Threshold above which solution is accepted with constant weight.
+                Defaults to 1e-1.
+            eta_2: Threshold above which solution is accepted with lower weight.
+                Defaults to 0.8.
+            ep: Threshold for virtual control weight update (nu > ep vs nu <= ep).
+                Defaults to 0.5.
+            eta_lambda: Step size for virtual control weight update. Defaults to 1e0.
+            lam_vc_max: Maximum virtual control penalty weight. Defaults to 1e5.
+            lam_prox_min: Minimum trust region (proximal) weight. Defaults to 1e-3.
+            lam_prox_max: Maximum trust region (proximal) weight. Defaults to 2e5.
+            lam_cost_drop: Iteration after which cost relaxation applies (-1 = never).
+                Defaults to -1.
+            lam_cost_relax: Factor applied to lam_cost after lam_cost_drop.
+                Defaults to 1.0.
         """
         self.rho_init = rho_init
         self.rho_max = rho_max


openscvx/algorithms/autotuning.py
```diff
@@ -1,24 +1,440 @@
...
+            state.accept_solution()
+            return "Accept Higher"
+
+
+def get_autotuner(settings: Config) -> AutotuningBase:
+    """Factory function to get the appropriate autotuning instance.
+
+    If a custom autotuner instance is provided in settings.scp.autotuner,
+    it will be used directly. Otherwise, defaults to AugmentedLagrangian()
+    with default parameters.
```
griffin-norris: I don't think we need this. Why do we need this factory at all, can't we just instantiate `self._autotuner = settings.scp.autotuner or AugmentedLagrangian()` inside `PenalizedTrustRegion.initialize()`?

openscvx/algorithms/penalized_trust_region.py
```diff
@@ -141,43 +185,89 @@ def step(
...
+            "J_tr": state.J_tr,
+            "J_vb": state.J_vb,
+            "J_vc": state.J_vc,
+            "cost": cost[-1],
+            "lam_prox": state.lam_prox,
+            "prob_stat": prob_stat,
+            "adaptive_state": adaptive_state,
+        }
+
+        # Only include nonlinear/reduction metrics when autotuner uses them
```
griffin-norris: I can come in and fix this once you have everything else done, but I think it would be nice if we could make our emitter slightly smarter so that we can turn up/down the verbosity etc. and have it nicely adapt the width and formatting to the number of elements

haynec: Agreed

openscvx/algorithms/base.py
```diff
@@ -82,9 +264,11 @@ def u(self) -> np.ndarray:
...
         return self.U[-1]
 
-    @property
-    def x_prop(self) -> np.ndarray:
-        """Extract propagated state trajectory from latest V.
+    def x_prop(self, index: int = -1) -> np.ndarray:
+        """Extract propagated state trajectory from V_history.
+
+        Args:
+            index: Index into V_history (default: -1 for latest entry)
```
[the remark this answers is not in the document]
haynec: Other autotuning methods make use of this as a function to grab the previous or earlier iterate discretization data, so I am in favor of keeping it a method.

haynec:
> @haynec I personally am not a huge fan of the `.candidate` pattern. Together with claude I drafted some problems/solutions
> 
> 1. Hidden state: The actual data (`inter_x`, `inter_u`, `inter_V`, _etc._) is stored as dynamically-set attributes on `AlgorithmState` via `setattr`, but none of these are declared in the dataclass. Anyone reading the class definition has no idea these attributes exist. This defeats the purpose of using a dataclass.
> 2. Fragile coupling: Adding a new candidate field requires adding a new property pair to `_IterateView` AND remembering the `inter_` prefix convention. Nothing enforces consistency.
> 3. Allocates on every access: `state.candidate` creates a new `_IterateView` instance every time it's called (it's a `@property`). Not a perf issue in practice for SCP loops, but needlessly wasteful.
> 4. `clear_candidate()` is manual and fragile: it has to null out each field individually (`base.py:248-258`). If someone adds a field to `_IterateView` but forgets to add it to `clear_candidate`, stale data persists silently.
> 
> As an alternative we could use a dataclass _e.g:_
> 
> ```python
> @dataclass
> class CandidateIterate:
>     x: Optional[np.ndarray] = None
>     u: Optional[np.ndarray] = None
>     V: Optional[np.ndarray] = None
>     x_prop: Optional[np.ndarray] = None
>     VC: Optional[np.ndarray] = None
>     TR: Optional[np.ndarray] = None
>     lam_vc: Optional[Union[float, np.ndarray]] = None
>     lam_cost: Optional[float] = None
>     lam_vb: Optional[float] = None
>     J_lin: Optional[float] = None
>     J_nonlin: Optional[float] = None
> ```
> 
> Then on `AlgorithmState`, just: `candidate: Optional[CandidateIterate] = None`. Clearing becomes `self.candidate = None` (or `self.candidate = CandidateIterate()`). The access pattern stays the same: `state.candidate.x`. It's ~15 lines instead of ~100, all fields are visible in one place, and type checkers can validate it.
> 
> I initially thought it would make sense to just store another `candidate: AlgorithmState` as a whole second state and just overwrite the candidate onto the active state but it's technically a slightly worse fit since the `AlgorithmState` carries a lot of state that the candidate doesn't need (notably the entire history, cost components, _etc._)

openscvx/algorithms/autotuning.py
```diff
@@ -1,24 +1,428 @@
...
-        scp_k: Current SCP iteration number
+        eta_0 = 1e-2
+        eta_1 = 1e-1
+        eta_2 = 0.8
+
+        gamma_1 = 2.0
+        gamma_2 = 0.5
+
+        lam_prox_min = 1e-3
+        lam_prox_max = 2e5
```
griffin-norris: Magic numbers! Can we add these to the `ScpConfig`?

openscvx/algorithms/autotuning.py
```diff
@@ -1,24 +1,428 @@
...
+
+        lam_prox_k = deepcopy(state.lam_prox)
```
griffin-norris: Isn't this just deepcopying a float?

openscvx/config.py
```diff
@@ -367,35 +366,109 @@ def __init__(
...
+        return self._autotuner
+
+    @autotuner.setter
+    def autotuner(self, value: Optional["AutotuningBase"]) -> None:
+        """Set a custom autotuner instance or reset to default when None.
+
+        Passing ``None`` clears the cached autotuner so that the next access
+        will recreate the default :class:`AugmentedLagrangian` instance.
+        """
+        self._autotuner = value
```
griffin-norris: I still don't understand why this needs to be a property, why can't we just have an `autotuner: AutotuningBase = AugmentedLagrangian` as part of the `ScpConfig`? So just a normal optional arg

openscvx/config.py
```diff
@@ -367,35 +366,109 @@ def __init__(
...
+            # Try to get n_states from direct reference first, then from parent config
+            n_states = None
+            if hasattr(self, "_n_states") and self._n_states is not None:
+                n_states = self._n_states
+            elif hasattr(self, "_parent_config") and self._parent_config is not None:
+                n_states = getattr(self._parent_config.sim, "n_states", None)
+
+            if self.n is not None and n_states is not None:
+                self._lam_vc = np.ones((self.n - 1, n_states)) * value
+                return
```
griffin-norris: This feels overengineered. As Claude and I understand it:

The setter tries to auto-expand scalars to arrays based on `_parent_config` and `_n_states`, which are set later in `Config.__post_init__`. This creates an ordering dependency: if someone sets `lam_vc` before the parent config is wired up, they get a scalar; if after, they get an array. This implicit behavior is fragile. Consider making the expansion explicit at the point of use (in `AlgorithmState.from_settings`) instead of hiding it in a setter.

I think the biggest thing that boths me is that we get EITHER an array OR a scalar which seems error prone and fragile. Is there a way we could keep this simpler?

openscvx/algorithms/autotuning.py
haynec: Yes, just did it out of an abundance of caution to set something in place.

openscvx/config.py
haynec: Basically, the actual weight is a matrix of `(N,n_x)`. So the simplest would be to have the user actually set the full matrix. For ease of use, I just made it so that the user can set it with a scalar, which is multiplied by a matrix of ones to set the actual weight.

openscvx/algorithms/autotuning.py
griffin-norris: could you try removing it and see if everything works? AFAIK floats are _never_ copied by python anyways so it shouldn't be an issue.

griffin-norris (approved):
LGTM.

There are some minor points of concern for me about how some parts of the code are organized but this is easy enough to refactor in the future as we progress in the development.

Make sure to squash-merge

PR #364, opened 2026-02-10
Added Support for nonlinear impulsive and hybrid systems

spadafab:
The present PR adds support to hybrid systems, i.e. allows to include both impulsive and continuous controls in the problem modeling and solution. Both linear and nonlinear impulsive controls are supported.

Large modifications are limited to the algorithm; minor modifications adapt symbolic and lowering scripts on the codebase side; adapt plotting tools on the preprocessing and postprocessing side.

Added free-final time hybrid double integrator to test correct working of the implemented changes. Plots follow.

newplot (1)

newplot

Added as well Hohmann transfer LEO to GEO as analytical solution comparison.

### DIFF
--- a/openscvx/symbolic/unified.py
+++ b/openscvx/symbolic/unified.py
@@ -269,11 +269,13 @@ def unify_controls(controls: List[Control], name: str = "unified_control") -> Un
     This function is the primary way to aggregate multiple symbolic Control objects into
     a single unified control vector for numerical optimization. It:
 
-    1. Sorts controls (user-defined first, augmented controls second)
-    2. Concatenates all control properties (bounds, guesses)
-    3. Assigns slices to each Control for extracting values from unified vector
-    4. Identifies special controls (time dilation)
-    5. Returns a UnifiedControl with all aggregated data
+    1. Reorders controls by actuation type: continuous first, impulsive second
+    2. Within each actuation-type group, places user-defined controls first and
+       augmented controls (names starting with ``"_"``) second
+    3. Concatenates all control properties (bounds, guesses)
+    4. Assigns slices to each Control for extracting values from unified vector
+    5. Identifies special controls (time dilation, impulsive masks/nodes)
+    6. Returns a UnifiedControl with all aggregated data
 
     Args:
         controls (List[Control]): List of Control objects to unify. Can include both
@@ -311,6 +313,25 @@ def unify_controls(controls: List[Control], name: str = "unified_control") -> Un
             print(unified._true_dim)         # 6 (thrust + torque)
             print(unified.true.shape)        # (6,)
             print(unified.augmented.shape)   # (1,) - time dilation
+
+        With both impulsive and augmented controls::
+
+            thrust = ox.Control("thrust", shape=(2,))  # continuous, user-defined
+            time_dilation = ox.Control("_time_dilation", shape=(1,))  # continuous, augmented
+            delta_v = ox.Control("delta_v", shape=(2,), impulsive=True, nodes=[0, 20])
+            delta_v_aug = ox.Control("_delta_v_bias", shape=(1,), impulsive=True, nodes=[0, 20])
+
+            unified = unify_controls([time_dilation, delta_v_aug, thrust, delta_v], name="u")
+
+            # Unified ordering is:
+            # [continuous true, continuous augmented | impulsive true, impulsive augmented]
+            # [thrust, _time_dilation | delta_v, _delta_v_bias]
+            print(thrust._slice)         # slice(0, 2)
+            print(time_dilation._slice)  # slice(2, 3)
+            print(delta_v._slice)        # slice(3, 5)
+            print(delta_v_aug._slice)    # slice(5, 6)
+            print(unified.slice_continuous)  # slice(0, 3)
+            print(unified.slice_impulsive)   # slice(3, 6)
 
     Note:
         After unification, each Control object has its `_slice` attribute set,


openscvx/algorithms/penalized_trust_region.py
```diff
@@ -343,14 +343,36 @@ def _subproblem(
...
+        idx_impulsive_controls = settings.sim.u.is_impulsive
+        idx_continuous_controls = ~settings.sim.u.is_impulsive
+
+        D_d = None
+
+        has_u_d = state.u[:, idx_impulsive_controls].size > 0
+        if has_u_d:
+            D_d_base = settings.sim.u.allocation_matrix
+            D_d = np.vstack((D_d_base, np.zeros((2, D_d_base.shape[1]))))
+
```
haynec: This should be incorporated into `augment_dynamics_with_ctcs`, https://github.com/OpenSCvx/OpenSCvx/blob/main/openscvx/symbolic/augmentation.py#L435-L589

examples/abstract/impulsive.py
```diff
@@ -0,0 +1,102 @@
...
+p.guess     = np.linspace(p.initial, p.final, n).reshape(-1, 1)
+
+v           = ox.State("velocity", shape=(1,))  # Scalar speed
+v.max       = np.array([1.0])
+v.min       = np.array([-1.0])
+v.initial   = np.array([0.0])
+v.final     = np.array([0.0])
+v.guess     = np.linspace(v.initial, v.final, n).reshape(-1, 1)
+
+dv           = ox.Control("delta_v", shape=(1,), impulsive=True, allocation_matrix=np.array([[0], [1]]), impulsive_nodes=[0, n-1] )
```
griffin-norris: As discusseed in person, it seems like we could derive the `allocation_matrix` from a `dynamics_impulsive` dict (mirroring the existing `dynamics` dict

This is a nitpick we could consider renaming `impulsive_nodes` to to just `nodes`, we know it's impulsive already based on `impulsive=True`. How do we nicely handle the fact that _if_ the control is impulsive _then_ it has these extra flags, otherwise it doesn't make sense to have them. If we have to manually guard against every invalid combination that could get tedious. Maybe we just capture them with a `*kwargs`?

openscvx/symbolic/expr/control.py
```diff
@@ -59,16 +59,41 @@ class Control(Variable):
...
+        ## TODO: (fabio) maybe this check could be moved to ''builder.py''
+        if impulsive:
+            if allocation_matrix.shape[1] != shape[0]:
+                raise ValueError(
+                    (
+                        "Allocation matrix dimensions not consistent with control dimensions.",
+                        "Number of rows shall be equal to control dimension.",
+                    )
+                )
+        self._allocation_matrix = allocation_matrix
```
griffin-norris: As discussed in person, it would be great to automate the construction of the allocation matrix from a `dynamics_impulsive` dict. In that case this would probably belong in `preprocessing.py` and get called from `builder.py`

openscvx/lowered/problem.py
```diff
@@ -26,6 +26,7 @@ class LoweredProblem:
 
     Attributes:
         dynamics: Optimization dynamics with fields f, A, B (JAX functions)
+        dynamics: Discrete optimization dynamics with fields f, A, B (JAX functions)
```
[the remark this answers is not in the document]
spadafab: right, addressed and corrected

openscvx/symbolic/expr/control.py
```diff
@@ -152,10 +163,46 @@ def scaling_max(self, val):
...
+            self._is_impulsive = np.repeat(False, self.shape)
+            return
+        val = np.repeat(val, self.shape)
+        if val.shape != self.shape:
+            raise ValueError(
+                (
+                    f"Impulsive controls toggles shape {val.shape} ",
+                    f" does not match Control shape {self.shape}",
+                )
+            )
```
griffin-norris: This returns a tuple of 2 strings, not one combined `str`
Should be:
```python
raise ValueError(
    f"Impulsive controls toggles shape {val.shape} "
    f"does not match Control shape {self.shape}"
)
```

openscvx/symbolic/preprocessing.py
```diff
@@ -799,6 +801,25 @@ def validate_input_types(
...
+    has_discrete_dynamics = dynamics_discrete is not None or has_discrete_from_byof
+    if any_impulsive_control_set and not has_discrete_dynamics:
+        raise ValueError(
+            "'dynamics_discrete' must be provided when at least one control is marked as "
+            "'impulsive'. Provide a dynamics_discrete dict and/or byof['dynamics_discrete']."
+        )
+    if (not any_impulsive_control_set) and dynamics_discrete is not None:
+        raise ValueError(
+            "'dynamics_discrete' must not be provided when no control is marked as 'impulsive'."
+        )
```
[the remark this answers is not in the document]
haynec: Yes

openscvx/symbolic/unified.py
```diff
@@ -325,11 +325,32 @@ def unify_controls(controls: List[Control], name: str = "unified_control") -> Un
...
+    sorted_controls = _stable_true_then_aug(controls_continuous) + _stable_true_then_aug(
+        controls_impulsive
+    )
+
+    # Assign slices in the new unified order.
+    offset = 0
+    for control in sorted_controls:
+        dim = int(control.shape[0])
+        control._slice = slice(offset, offset + dim)
+        offset += dim
```
griffin-norris: We should update the docstring of this function to reflect the changed behavior. I would specifically want to be able to understand the ordering in the case where we have both impulsive and augmented controls. This can probably be accomplished with an example in the docstring

griffin-norris (changes requested):
Looks very good, much much better than the first draft. A few minor changes requested. The biggest issue is the missing unit test

openscvx/symbolic/unified.py
spadafab: yep that is on me

haynec (approved):
LGTM assuming the doc changes are done per @griffin-norris comment

PR #369, opened 2026-02-12
Refactor/discretization base class

### LINKED ISSUE #122: Improve Documentation
- Add docstrings
- Improve type hinting to be more accurate
- Add more examples

### LINKED ISSUE #127: Add Documentation for Discretization

### LINKED ISSUE #200: Add Base Classes to Allow Users to Provide Algorithms
- Add base class for **discretization**, convert existing (purely functional) discretization logic into an instance of the class
- Add base class for **SCP algorithm**, convert existing algorithm  into an instance of the class
- Add base class for **convex subproblem**, convert existing convex subproblem into an instance of the class
- Add base class for **integration**, convert existing integration logic into an instance of the class

### LINKED ISSUE #301: Add Base Class for Discretization

griffin-norris:
- Combines linearization and discretization into a single `Discretizer` ABC -> Fixes #301 -> Fixes #200, Fixes #127 -> Fixes #122
- Different schemes may linearize-then-discretize or discretize-then-linearize, so a unified interface keeps the contract consistent (continuous nonlinear dynamics in, discrete linear matrices out)
- Jacobian computation (`jax.jacfwd`) moves into the `discretizer`,  `Dynamics.A/B` are no longer set during lowering or used internally, simplifying BYOF and CTCS code paths
- Existing behavior preserved as `LinearizeDiscretize(Discretizer)` with full citations
- Follows the same ABC pattern as `Algorithm` and `ConvexSolver`: setup-once via `get_solver()`, per-iteration via the returned callable, plus `citation()`

```python
class Discretizer(ABC):
    """Convert continuous-time nonlinear dynamics into discrete-time
    linear approximations around a reference trajectory."""

    @abstractmethod
    def get_solver(self, dynamics: Dynamics, settings: Config) -> callable:
        """Return a callable (x, u, params) -> (A_d, B_d, C_d, x_prop, V)."""
        ...

    @abstractmethod
    def citation(self) -> List[str]:
        """Return BibTeX citations for this discretization method."""
          ...
```

### DIFF
--- a/openscvx/algorithms/penalized_trust_region.py
+++ b/openscvx/algorithms/penalized_trust_region.py
@@ -21,9 +21,9 @@ from openscvx.utils.printing import (
     color_prob_stat,
 )
 
+from .base import Algorithm, AlgorithmState, CandidateIterate
 from .ConstantProximalWeight import ConstantProximalWeight
 from .RampProximalWeight import RampProximalWeight
-from .base import Algorithm, AlgorithmState, CandidateIterate
 
 if TYPE_CHECKING:
     from openscvx.lowered import LoweredJaxConstraints


haynec (changes requested):
Looks good for the most part, I think we should move ```dVdt()``` to live in the LinearizeDiscretize class.

PR #379, opened 2026-02-17
Feature/sparsity calculation

### LINKED ISSUE #341: Sparse DPP parameters for CVXPY
Convex subproblem solves can be made substantially faster if `cp.Parameter`s associated with linearization terms (dynamics + nodal constraints) are given a known sparsity pattern to reduce the overall parameter count. e.g. `Ak = cp.Parameter((nx,nx), sparsity=sparsity)` for the DT-LTV system dynamics matrix. If no sparsity pattern is supplied, then e.g. `Ak` scales `O(nx^2)` in terms of parameter count, leading to slower solve times as the state-space dimensionality scales.

The rough necessary steps to do this include:
1. Ahead-of-time: Derive, from the symbolic expression tree (implemented) of dynamics/nodal constraints, the sparsity pattern associated with each linearization term.
2. Ahead-of-time: Convert to boolean matrix where `1` indicates non-zero entry.
3. Ahead-of-time: Construct matrix- or vector-valued parameter as `P = (cp.Parameter(,...), sparsity=sparsity)`
4. Run-time: populate `P.value_sparse` with a `scipy.sparse.coo_array` object

More info here: [https://www.cvxpy.org/tutorial/constraints/index.html](https://www.cvxpy.org/tutorial/constraints/index.html)

griffin-norris:
- Add static sparsity analysis via `Expr.sparsity(n_x, n_u)`, wired into CVXPY's sparsity parameter -> Fixes #341
  - Element-level boolean Jacobian patterns are propagated through the AST using chain/product rule semantics (no JAX evaluation needed)
  - Liveness gating in `Mul`/`Div`/`Power` eliminates false positives when constant zeros are present (e.g. sparse rotation matrices, zero exponents)
  - `Expr` provides a conservative default `.sparsity(...)`
- `transitive_closure` converts continuous-time patterns to discrete-time (`A_d`, `B_d`, `C_d`)

```python
# 3-DOF rocket: pos(3), vel(3), mass(1), thrust(3)
dynamics = ox.Concat(
    vel,                         # pos_dot = vel
    thrust / mass - g,           # vel_dot = thrust/mass - g
    -alpha * ox.Norm(thrust),    # mass_dot = -α‖T‖
)

A_c, B_c = dynamics.sparsity(n_x=7, n_u=3)

#        pos  vel  mass          thrust
# A_c = [  .   I    .  ]  B_c = [  .  ]   ← pos_dot depends only on vel
#        [  .   .    x  ]        [  I  ]   ← vel_dot depends on mass & thrust
#        [  .   .    .  ]        [ xxx ]   ← mass_dot depends on all thrust (Norm)

# Lift to discrete-time via transitive closure of A_c
A_d, B_d, C_d = ox.discrete_sparsity(A_c, B_c, dis_type="FOH")

# A_d now shows pos → mass coupling (indirect, through vel)
assert A_d[0, 6] == True   # pos[0] depends on mass after discretization
```

> [!NOTE]
> Sparsity _can_ vastly improve performance (we measured up to 100x) in favorable conditions but it is not _guaranteed_ to do so. It is most applicable to very large, sparse problems where we can vastly reduce the number of zeros

> [!WARNING]
> There is an [issue](https://github.com/cvxpy/cvxpy/issues/3145) in CVXPY `v1.8.1` preventing the sparsity argument for `cp.Parameter` from having any effect. They have merged a fix in [PR #3146](https://github.com/cvxpy/cvxpy/pull/3146), until that is released you must locally install CVXPY to see any speedup

### DIFF
--- a/openscvx/solvers/base.py
+++ b/openscvx/solvers/base.py
@@ -30,7 +30,7 @@ must follow for use within successive convexification algorithms.
 """
 
 from abc import ABC, abstractmethod
-from typing import TYPE_CHECKING, Any, List
+from typing import TYPE_CHECKING, Any, List, Optional
 
 if TYPE_CHECKING:
     from openscvx.config import Config
@@ -94,8 +94,8 @@ class ConvexSolver(ABC):
         x_unified: "UnifiedState",
         u_unified: "UnifiedControl",
         jax_constraints: "LoweredJaxConstraints",
-        dynamics_sparsity=None,
-        constraint_sparsity=None,
+        dynamics_sparsity: Optional[tuple] = None,
+        constraint_sparsity: Optional[list] = None,
     ) -> None:
         """Create backend-specific optimization variables.
 
--- a/openscvx/solvers/ptr_solver.py
+++ b/openscvx/solvers/ptr_solver.py
@@ -7,7 +7,7 @@ code generation via cvxpygen for improved performance.
 
 import os
 from dataclasses import dataclass
-from typing import TYPE_CHECKING, List
+from typing import TYPE_CHECKING, List, Optional
 
 import cvxpy as cp
 import numpy as np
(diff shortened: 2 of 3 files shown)


haynec (approved):
LGTM

PR #402, opened 2026-02-27
Chore/cleanup impulsive

griffin-norris:
- Minor naming tweaks
- Move `create_hohmann_transfer_server` to `examples/plotting_viser.py`
- Decrease unit-test `dt` for Hohmann transfer problems to speed up computation
- Change `show_grid` to default `False` for Hohmann transfer visualization


haynec (approved):
LGTM

PR #414, opened 2026-03-04
Feature/disclin

haynec:
Changelog:
-Adds `ox.DiscretizeLinearize` to discretization methods.
-Fixes https://github.com/OpenSCvx/OpenSCvx/issues/406

### DIFF
--- a/openscvx/discretization/discretize_linearize.py
+++ b/openscvx/discretization/discretize_linearize.py
@@ -185,7 +185,19 @@ class VectorizeDiscretizeLinearize(Discretizer):
         return solver
 
     def citation(self) -> List[str]:
-        return []
+        """Return BibTeX citations for the vectorize-then-discretize-then-linearize method.
+
+        Returns:
+            List containing the BibTeX entries
+        """
+        return [
+            r"""@phdthesis{kidger2021on,
+  title={{O}n {N}eural {D}ifferential {E}quations},
+  author={Patrick Kidger},
+  year={2021},
+  school={University of Oxford},
+}""",
+        ]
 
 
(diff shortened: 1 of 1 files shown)


haynec:
> > > I've noticed that `VectorizeDiscretizeLinearize` is generally a second or two slower to compile than `LinearizeDiscretize` on my laptop. Maybe I've implemented some JAX stuff inefficiently. Still, failing the brachistochrone initialization speed by 0.06 seconds is just adding insult to injury :P
> > 
> > 
> > > 
> > 
> > 
> > > Is there a way to see how fast the python3.12/Ubuntu VM initializes the brachistochrone problem with `LinearizeDiscretize`? It'd be helpful to know if it's teetering on the edge of the 10 s limit and this just pushed it over, or if my implementation is horribly increasing the compile time.
> > 
> > 
> > Failing by 0.06 is close enough. We can just relax that timing result in the testing. Solve time is much more interesting then compilation time.
> 
> I think we should at least investigate it. While compilation time isn't critical, I still wouldn't want it to continually slip.
> 
> This seems like a classic case of [_normalization of deviance_](https://en.wikipedia.org/wiki/Normalization_of_deviance)

@griffin-norris I would argue that the timing results for examples are only relevant when performed on the github actions on github hardware and not locally. Otherwise hardware differences will obscure algorithmic changes which cause timing results to change. 

Also the timing bounds for compilation is something we pulled out of thin air once based on the performance we happened to be getting at that time, it is not something I've ever put any great thought into choosing. It's important if something slows down compilation by a large amount, meaning on the order of minutes. I would be supportive of significantly relaxing the timing bounds for compilation since in my mind that check should be for if compilation time is significantly slowed, not marginally.

PR #452, opened 2026-04-08
Feature/add remaining stl operators

### LINKED ISSUE #373: Add Remaining STL Operators
- Add remaining STL operators: `And`, `Always`, `Eventually`, `Until`
- Should/Could probably leave the intervals out for now until #325 is merged since we need to be careful about wall-time and nodes for intervals
- Also need to add unit tests for the STL operators

griffin-norris:
## Summary

- Added remaining STL operators -> Fixes #373
    - Added `Not` STLExpr (with double-negation elimination)
    - Promoted `Always` to a real `STLExpr` node (was: function returning `CTCS`) so it nests inside `And`/`Or`/`Not`/
`IfThen` and composes with the new operators. `.over()` still seals it into a `CTCS`.
    - Added `Eventually` and `Until` placeholder nodes — constructible (so downstream code can be sketched) but raise `NotImplementedError` on lower/enforce. Real formulation in progress.
- Added `&`/`|`/`~` operator overloads on `STLExpr` so STL specs can be composed with natural syntax. Operators are intentionally **not** added to `Constraint` to avoid colliding with `logic.py`'s hard-boolean `All`/`Any`/`Cond`.

## New surface

```python
import openscvx as ox

avoid  = ox.linalg.Norm(position - obs)    >= safety_radius
in_box = ox.linalg.Norm(position - center) <= box_radius

# Standalone — interval at construction, .over() seals into a CTCS
constraints.append(ox.stl.Always(avoid, (0, N - 1)).over())

# Nested composition with operator sugar
spec = (ox.stl.Always(avoid) & ox.stl.Always(in_box)) | ~stuck
constraints.append(spec.over((0, N - 1)))
```

> [!NOTE]
> Intervals currently live in exactly one place: on the operator or at `.over()`
> ```python
> Always(p, I).over(...)  # raises
> Always(p, I).at(...)    # raises
> ```
> This may change once we properly support intervals for the STL operators

### DIFF
--- a/openscvx/symbolic/expr/__init__.py
+++ b/openscvx/symbolic/expr/__init__.py
@@ -65,8 +65,11 @@ Module Organization:
         `NodalConstraint` for enforcing constraints at discrete nodes and `CTCS` for
         continuous-time constraint satisfaction.
 
-    Signal Temporal Logic (stljax.py):
-        `Or` for logical disjunction in task specifications (stljax-backed).
+    Signal Temporal Logic (stl.py):
+        GMSR-backed smooth task-logic operators (`And`, `Or`, `Not`, `IfThen`,
+        `Always`, `Eventually`, `Until`, `IntegerVariable`) accessible via the
+        `ox.stl` namespace. See also `logic.py` for hard-boolean branching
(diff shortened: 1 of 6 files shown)


haynec (approved):
LGTM

PR #454, opened 2026-04-12
updated plot_camera_animation for the current results version

haynec:
Changelog:
- Updated plot_camera_animation for the current results version


griffin-norris (approved):
LGTM

PR #458, opened 2026-04-13
Feature/optional solver time limit

### LINKED ISSUE #386: Option to Limit Iterations by Wall-clock
- Would it make sense to allow users to provide a total wall clock time limit rather than or in adition to `k_max`?
- Would be useful for MPC use cases with strict time budgets

griffin-norris:
- Add optional wall-clock time limit (`t_max`) for SCP solves, symmetrical to `k_max` -> Fixes #386
- Default is None (off) — no behavior change for existing users
- Configurable via algorithm dict, algorithm instance, or per-solve override
- Time is checked between iterations (never interrupts a subproblem mid-solve)
- Solve reports `converged=False` on timeout, same as hitting the iteration limit

```python
# Algorithm config (same pattern as k_max)
problem = Problem(..., algorithm={"k_max": 200, "t_max": 30.0})

# Per-solve override
result = problem.solve(time_limit=10.0)
```

### DIFF
--- a/openscvx/algorithms/__init__.py
+++ b/openscvx/algorithms/__init__.py
@@ -129,6 +129,7 @@ class PenalizedTrustRegionConfig(BaseModel):
         return v
 
     k_max: int = 200
+    t_max: Optional[float] = None
     lam_prox: Union[float, Dict[str, Any]] = 1e-1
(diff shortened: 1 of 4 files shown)


haynec (approved):
LGTM

PR #462, opened 2026-04-14
Chore/manipulation self collisions

griffin-norris:
- Adds `examples/arm/7_dof_arm_collision.py`: a variant of `7_dof_arm.py` that enforces **capsule–capsule self-collision** between non-adjacent links and **capsule–sphere avoidance** against a spherical obstacle. Both are implemented as continuous-time (`ox.ctcs`) constraints, batched with `ox.Vmap` over sample-pair grids along each link.
- Reuses the existing `joint_transforms` snapshots from the FK chain to lift keypoints into the world frame — no new kinematic machinery.
- Extends `docs/UsersGuide/07_lie.md` with an **Extension: Self-Collision and Obstacle Avoidance** section covering the capsule approximation, the non-adjacent-pair skip, and the obstacle variant. Adds a `!!! warning` callout noting the current solve-time regression and links the new example from the Further Reading block.

### DIFF
--- a/docs/UsersGuide/07_lie.md
+++ b/docs/UsersGuide/07_lie.md
@@ -364,6 +364,98 @@ results = problem.post_process()
 
 After solving, `results.trajectory["T_ee"]` gives you the dense end-effector transform along the propagated trajectory, ready to feed straight into a viser animation or a downstream controller.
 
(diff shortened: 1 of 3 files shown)


haynec (approved):
LGTM

PR #471, opened 2026-05-07
Fix spelling of 'initialise' to 'initialize'

haynec:
Corrected spelling of 'initialise' to 'initialize' in README.


griffin-norris (approved):
Tsk tsk tsk

PR #474, opened 2026-05-11
Feature/lazy variable guess initialization

### LINKED ISSUE #473: Custom time grid
It would be nice to feature a time grid guess that is different from uniform among a list of available and validated ones.

A preliminary version is in the current draft PR `https://github.com/OpenSCvx/OpenSCvx/pull/472`, from branch `https://github.com/OpenSCvx/OpenSCvx/tree/feature/customized_time_grid`.

The front-end should mirror the current implementation of `hypersensitive.py` found in the above PR and branch.

Retrieved from https://github.com/purnanandelango/trajopt-util/tree/master/mutil/%2Bgrid

griffin-norris:
## Summary

- `.guess` continues to accept arrays for the common case — the existing eager form is unchanged. Callables are an additive opt-in for cases that need lazy evaluation: N-agnostic problem definitions, cross-variable dependencies, or MPC warm-starts. Fixes #473
- When provided, callable parameter names dispatch automatically:
  - The reserved name `tau` receives a normalized `[0, 1]` grid of length `N`
  - Any other parameter is matched against a state/control in the problem and receives that `Variable` (read `.guess` for the resolved array).
- Resolution is topo-sorted with cycle detection, runs in `resolve_guesses(states, controls, N)` before augmentation, and re-runs on every `Problem.sync` / `solve` / `reset` — closures over `state.initial` reflect mutations live.
- Fallback defaults (linear `initial→final` for states with both, broadcast `initial` for propagation-only accumulators, finite-difference dt/dtau for `time_dilation_guess`) are themselves callables — one resolution mechanism, no separate code path.
- Callable signatures are inspected eagerly at assignment, so `*args` / `**kwargs` / positional-only get rejected pointing at the user's lambda, not at preprocessing.

## Usage

Arrays still work exactly as before — most users won't need to change anything:

```python
pos.guess = np.linspace([0, 0, 0], [10, 5, 2], N)
```

The lazy form is there when you need it. Distribution-driven time guess — e.g. cosine clustering at the endpoints for a hypersensitive problem (see #472):

```python
time = ox.Time(
    initial=0.0,
    final=1000.0,
    min=0.0,
    max=1000.0,
    time_dilation_min=1e-4,
    time_dilation_max=1e4,
    guess=lambda tau: 1000.0 * 0.5 * (1.0 - np.cos(np.pi * tau)),
)
```

No `time_dilation_guess` needed — the default closure finite-differences `time.guess` on every resolve.

Cross-variable guesses dispatched by parameter name:

```python
pos = ox.State("pos", shape=(3,))
pos.initial = [0.0, 0.0, 0.0]
pos.final   = [10.0, 5.0, 2.0]
# Shape-only: tau is the normalized [0, 1] grid of length N.
pos.guess   = lambda tau: np.outer(tau, [10.0, 5.0, 2.0])

vel = ox.State("vel", shape=(3,))
# Cross-var: `pos` is the State object, resolved before `vel`. Read .guess for the array.
vel.guess   = lambda pos: np.gradient(pos.guess, axis=0)

thrust = ox.Control("thrust", shape=(3,))
# Multi-var: F = m * dv/dt — `time` is the Time state, dispatched the same way.
thrust.guess = lambda vel, time: m * np.gradient(vel.guess, time.guess[:, 0], axis=0)
```

MPC warm-start — closure captures the State, so mutating `pos.initial` between solves flows through automatically:

```python
pos.guess = lambda tau, _self=pos: np.linspace(
    _self.initial, _self.final, len(tau)
)

# Between solves:
pos.initial = measured_state
problem.sync()   # re-resolves pos.guess against the new initial
```

## Out of scope

- YAML/dict serialization of callables (`VariableSpec` / `TimeSpec` still accept arrays only).

### DIFF
--- a/openscvx/problem.py
+++ b/openscvx/problem.py
@@ -516,6 +516,11 @@ class Problem:
         workflows where the initial trajectory guess is updated between solves (e.g., shifting
         the previous solution).
 
+        Any state or control whose guess was set as a callable is re-resolved
+        here, so that callables which capture mutable state (e.g.
+        ``lambda tau: state.initial + ...``) reflect the current problem
+        configuration on every sync.
+
         Note:
             This only updates the unified representation. The AlgorithmState is
             created from these values in reset() or initialize(), so this must
@@ -523,6 +528,13 @@ class Problem:
         """
         if self._lowered is None:
             return
+
+        from openscvx.symbolic.preprocessing import resolve_guesses
+
+        N = self.symbolic.N
+        # Re-resolve callable guesses against the current N. States/controls
+        # whose guess is an explicit array are left untouched.
+        resolve_guesses(self.symbolic.states, self.symbolic.controls, N)
 
         # Sync optimization state guesses
         for state in self.symbolic.states:
--- a/openscvx/problem.py
+++ b/openscvx/problem.py
@@ -532,9 +532,11 @@ class Problem:
         from openscvx.symbolic.preprocessing import resolve_guesses
 
         N = self.symbolic.N
-        # Re-resolve callable guesses against the current N. States/controls
-        # whose guess is an explicit array are left untouched.
-        resolve_guesses(self.symbolic.states, self.symbolic.controls, N)
+        # Re-resolve callable guesses against the current N. ``states_prop``
+        # is a superset of ``states`` (includes propagation-only extras), so
+        # this covers opt and prop in one pass. States/controls whose guess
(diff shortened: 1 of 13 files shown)


spadafab:
The code looks clean and a basis for potential further developments. The current code looks aligned with the feature request #473 and, except for the only comment, it looks aligned with the contents from #472

PR #483, opened 2026-05-14
block xla caching in init until ready

haynec:
Ensures that all JAX/XLA compilation steps are performed in `problem.initialize()`. This reduces the overhead in the `problem.solve()`. This took the main solve time for `6DoF_pdg.py` from 53ms to 12ms on an MBP M2Max.

### DIFF
--- a/openscvx/algorithms/penalized_trust_region.py
+++ b/openscvx/algorithms/penalized_trust_region.py
@@ -6,8 +6,9 @@ optimization problems through iterative convex approximation.
 
 import time
 import warnings
-from typing import TYPE_CHECKING, Callable, Dict, List, Union
+from typing import TYPE_CHECKING, Callable, Dict, List, Tuple, Union
 
+import jax
 import numpy as np
 import numpy.linalg as la
 
@@ -188,6 +189,15 @@ class PenalizedTrustRegion(Algorithm):
             return solver.call(*args)
         return solver(*args)
 
+    @staticmethod
+    def _block_until_ready_outputs(outputs: Tuple[object, ...]) -> None:
+        """Finish any pending XLA work from discretization exports (warm-up helper)."""
+        try:
+            jax.block_until_ready(outputs)
+        except (TypeError, ValueError):
+            # Non-JAX leaves (e.g. pure NumPy); nothing to wait on.
+            pass
+
     def _recover_prior_node_from_initial(
         self,
         settings: Config,
@@ -286,7 +296,8 @@ class PenalizedTrustRegion(Algorithm):
         """Initialize PTR algorithm.
 
         Stores compiled infrastructure and performs a warm-start solve to
-        initialize DPP and JAX jacobians.
+        initialize DPP and JAX jacobians. Also runs post-subproblem discretization
+        on the throwaway CVX solution so XLA/export caches match the first ``step()``.
 
         Args:
             solver: Convex subproblem solver (e.g., CVXPySolver)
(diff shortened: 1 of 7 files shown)


openscvx/algorithms/penalized_trust_region.py
```diff
@@ -333,7 +344,41 @@ def initialize(
...
+            _costs,
+            _result_cost,
+            _J_vb_vec,
+            _J_vc_vec,
+            _J_tr_vec,
+            _prob_stat,
+            _subprop_time,
+            _vc_mat,
+            _tr_mat,
+        ) = self._subproblem(params, init_state, settings)
```
griffin-norris: Could this be `x_sol, u_xol, *_ = self._subproblem(...)`

openscvx/algorithms/penalized_trust_region.py
```diff
@@ -333,7 +344,41 @@ def initialize(
...
+        # initial guess, so the first step() in solve() could still hit an XLA cache_miss
+        # on post-CVX (x_sol, u_sol). Running that path here moves compilation into init.
+        cont_out = self._invoke_solver(
+            self._discretization_solver, x_sol, u_sol.astype(float), params
+        )
+        x_prop_c = cont_out[3]
+        u_candidate = u_sol.astype(float)
+        x0_prior_c = self._recover_prior_node_from_initial(settings, x_sol[0])
+        x_nodes_prior_c = np.vstack((x0_prior_c, np.asarray(x_prop_c)))
+        if self._discretization_solver_impulsive is not None:
```
griffin-norris, line 356: The other impulsive guard is `if has_impulsive and self_discretization_solver_impulsive is not None`. Claude is concerned that this guard is different here

openscvx/algorithms/penalized_trust_region.py
```diff
@@ -188,6 +189,15 @@ def _invoke_solver(solver: callable, *args):
             return solver.call(*args)
         return solver(*args)
 
+    @staticmethod
+    def _block_until_ready_outputs(outputs: Tuple[object, ...]) -> None:
+        """Finish any pending XLA work from discretization exports (warm-up helper)."""
+        try:
+            jax.block_until_ready(outputs)
+        except (TypeError, ValueError):
+            # Non-JAX leaves (e.g. pure NumPy); nothing to wait on.
```
griffin-norris: Do we have non-jax leaves that hit here? Nothing gets lowered to NumPy, right?

griffin-norris (commented):
Mostly looks good, a few minor nitpicks. We should for sure address the test issue before merging.

Do we know why we get the cache miss?

openscvx/algorithms/penalized_trust_region.py
haynec: Addressed

openscvx/algorithms/penalized_trust_region.py
haynec: Changing them to be the same seems to reintroduce this overhead. Claude summarized it with this.

First block (with has_impulsive) exists so init_state only gets add_impulsive_discretization(...) when the problem actually has an impulsive control slice. That is about correct warm-start state, not a micro-optimization.

Second block (cache priming after the dummy subproblem) is meant to mirror what step() does after every subproblem: it always calls the impulsive discretization when that solver is present, without a has_impulsive check. So the looser guard keeps init priming aligned with the real iteration path and avoids moving XLA/compile cost to the first step() instead.

has_impulsive “overhead” is trivial (a slice comparison). The real cost is an extra _invoke_solver impulsive call; skipping it only helps if you also change step() (and the state.k == 1 path) the same way, otherwise behavior and compile timing diverge.

openscvx/algorithms/penalized_trust_region.py
haynec: Addressed

griffin-norris (approved):
LGTM

PR #489, opened 2026-05-16
Feature/qpax convex backend

### LINKED ISSUE #27: Add differentiable solver using qpax
- Use [qpax](https://github.com/kevin-tracy/qpax) to formulate the optimal control problem
- Necessitates concatenation logic to form single QP for the problem

griffin-norris:
## Summary

- Adds **`QPAXPTRSolver`** — a JAX-native QP backend for the PTR convex subproblem, sitting alongside `CVXPyPTRSolver` behind a new abstract `PTRSolver` base. Selected via `solver={"backend": "qpax"}`. Optional install: `pip install openscvx[qpax]`. → Fixes #27
- **Why now:** wedge toward end-to-end JAX differentiability — `qpax.solve_qp_primal` is `jax.custom_vjp`-differentiable, so once the surrounding pipeline (discretizer, algorithm loop, parameter sync) becomes JIT-friendly in a follow-up, `jax.vmap` / `jax.grad` will reach through a full SCvx solve. This PR is only the backend.
- **No behavior change on the default CVXPy path.** `ox.PTRSolver` is now the abstract base; the previous concrete class is `ox.CVXPyPTRSolver`. `SolverSpec` is kept as a deprecated alias of `PTRSolverSpec`.

## Notable design

- **Per-backend convex lowering.** `ConvexSolver.lower_convex_constraints` is the new seam — default refuses user `.convex()`, `CVXPyPTRSolver` overrides. No more solver-type branching in `symbolic/lower.py`.
- **`PTRSolverSpec.backend` discriminator** with a model validator that rejects CVXPy-only fields (`cvx_solver`, `cvxpygen`) under `backend="qpax"`.
- **CTCS supported in v1** (scope expansion vs original plan — the library auto-adds time-bound CTCS to every problem, so a hard reject would have rejected every problem).
- **Rejected in v1, with clear "use `CVXPyPTRSolver`" errors:** user `.convex()`, cross-node constraints, impulsive controls.

## Known v1 gaps (called out in docstrings, follow-up material)

- **No warm-start** — `qpax.solve_qp` initializes its own primal-dual state with no init hook (only `relax_qp` accepts one).
- **Per-iteration JIT-compilation cost** on first call to `qpax.solve_qp`; full JIT of `_assemble_qp` is the follow-up that turns this into "differentiable SCvx."

## Test plan

- [x] `tests/solvers/test_qpax_ptr_solver.py` — 7 new unit tests: ABC contract, install guard, spec validation, `.convex()` reject, round-trip vs CVXPy on a double-integrator, `_assemble_qp` shape sanity, `PTRSolveResult` return.
- [x] `tests/test_brachistochrone.py::test_backend[cvxpy|qpax]` — parametrized analytical-cycloid check; both backends hit the same accuracy thresholds.
- [x] Full suite green: `pytest -n auto -m 'not integration'` → **1770/1770 passing**.

## Files of note

- `openscvx/solvers/ptr_solver.py` — new abstract `PTRSolver` (ABC) + `PTRSolveResult`
- `openscvx/solvers/cvxpy_ptr_solver.py` — renamed concrete impl, unchanged behavior
- `openscvx/solvers/qpax_ptr_solver.py` — new backend
- `openscvx/solvers/base.py` — `PTRSolverSpec` discriminator, default `lower_convex_constraints`
- `openscvx/symbolic/lower.py` — backend-agnostic delegation
- `docs/UnderTheHood/lowering_architecture.md` — new "Convex Subproblem Backends" section

### DIFF
--- a/docs/UnderTheHood/lowering_architecture.md
+++ b/docs/UnderTheHood/lowering_architecture.md
@@ -160,9 +160,40 @@ JAX lowering has no dependency on:
 
 This means JAX-lowered dynamics and constraints could be used with alternative solvers.
 
+## Convex Subproblem Backends
+
+The convex subproblem at each SCP iteration is solved by a concrete subclass
+of the abstract `PTRSolver` base. Two backends ship today:
+
+| Backend | Class | Selector | Notes |
+|---------|-------|----------|-------|
+| CVXPy (default) | `CVXPyPTRSolver` | `solver={"backend": "cvxpy"}` (default) | DCP graph via CVXPy, dispatched to QOCO / CLARABEL / etc. Supports user `.convex()` constraints, cross-node constraints, CTCS, and impulsive controls. Optional cvxpygen code generation. |
+| QPAX | `QPAXPTRSolver` | `solver={"backend": "qpax"}` | JAX-native QP via `qpax.solve_qp`. Flat `(Q, q, A, b, G, h)` assembly. Supports box / dynamics / CTCS / boundary-Fix; **rejects** user `.convex()`, cross-node, and impulsive at `initialize()` with a clear "use `CVXPyPTRSolver`" message. Enables a path toward an end-to-end JAX-differentiable SCP loop in follow-up work. |
+
+Picking a backend at construction time:
+
+```python
+import openscvx as ox
+
+# Default — same as today's CVXPyPTRSolver.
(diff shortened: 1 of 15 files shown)


haynec (approved):
Assuming you are ok with the GitHub quality check then this LGTM

Issue #494, opened 2026-05-16
Support for Batchable/Differentiable Solves


griffin-norris:

griffin-norris:
The batchable half of this issue shipped across the same PR line as #493 — #514 (`solve_jax`), #520 (`solve_batched`), #525 (unified batching API), enabled by #501, #504, #512, #529, #530.

Descoping the differentiable half for now and closing: #512 deliberately moved QPAX off the `custom_vjp` `solve_qp_primal` path to get a reliable convergence flag, Moreau's implicit-differentiation hook is still future work, CVXPy's `pure_callback` is non-differentiable by construction, and reverse mode through the `lax.while_loop` solve loop would need a fixed-iteration `scan` variant or implicit differentiation at the converged point. There is currently no `jax.grad`-through-solve test coverage, and the docs describe `jax.grad` as best-effort/untested.

If grad-through-solve comes back on the table, it deserves a fresh, scoped issue covering: flipping QPAX to the differentiable path (while keeping a convergence flag), threading Moreau's KKT implicit differentiation through the SCP loop, and a finite-difference validation test of `jax.grad(loss)(x0)` through `solve_jax`.

Issue #493, opened 2026-05-16
Make Entire Solve Call Batchable


griffin-norris:

griffin-norris:
Resolved retroactively by the batchable-solve line: #514 (`Problem.solve_jax()` — the entire solve call composes with `jax.vmap`/`jax.jit`), #520 (`Problem.solve_batched()`), and #525 (unified rank-based batching API), enabled by #501, #504, #512, #529, and #530. All three entry points are covered by e2e vmap/batched tests and shipped batched examples.

PR #501, opened 2026-05-17
Refactor/jax traceable autotuners

griffin-norris:
Rewrites every autotuner's `update_weights` as a pure functional update on a JAX-traceable `AlgorithmState` pytree, so any autotuner composes with `jax.vmap` / `jax.jit` / `jax.grad`. Implementation plan: `plans/jax-traceable-autotuners.md`.

## Changes

- **`AlgorithmState`** is now a frozen, `jax.tree_util.register_pytree_node_class` dataclass holding only the current iterate (18 jnp leaves). Append-only history (X, U, discretizations, `lam_*`, `J_*`, pred/actual/acceptance) moves to a new CPU-side `AlgorithmHistory`, populated by the SCP loop via `record_iteration`.
- **`AutotuningBase.update_weights`** now returns `AlgorithmState` instead of a string; the four-bucket acceptance-ratio decision is encoded as an `AdaptiveStateCode` IntEnum on `state.adaptive_state_code`. `calculate_cost_from_state` / `calculate_nonlinear_penalty` are jnp-only and trace under `jax.jit`.
- **All four autotuners ported.** AL and APW use `jax.lax.cond` on `k == 1` plus a `jnp.where` cascade over the rho buckets. The `predicted_reduction == 0` case is now a deterministic `rho = -inf` (falls into reject) instead of a `ValueError`. Constant/Ramp are straight functional replacements.
- **`PenalizedTrustRegion.step`** now takes `(state, history, params, settings)` and returns `(next_state, converged)`. `Problem` threads `AlgorithmHistory` alongside `_state` through every public entry point; `Problem.history` exposes the log for diagnostics.

## Verification

- `tests/test_brachistochrone.py::test_autotuning` (parametrized across all four autotuners) — passes; the analytic-cycloid comparison is the load-bearing acceptance gate.
- New `tests/algorithms/test_algorithm_state_pytree.py` — pytree round-trip + `replace()` semantics (6 tests).
- New `tests/algorithms/autotuner/test_update_weights_jit.py` — `jax.jit(update_weights)` returns the same pytree as the bare call for each autotuner on both `k=1` and `k>1` branches (6 tests).
- `tests/test_autotuning.py` ported from 1743 → 1098 lines; 36 semantic survivors pass, 6 mutate-in-place tests deleted.
- Full suite (`pytest -m 'not integration'`) passes; the only parallel-mode flakes are timing-budget violations on brachistochrone solves under contention — they resolve serially.

## Out of scope

- The SCP outer loop staying Python — `lax.while_loop` migration is the next plan (`plans/batchable-problem.md`).
- `CandidateIterate` becoming a registered pytree — deferred until the discretizer/subproblem path is JAX-pure end-to-end.
- One sharp edge worth a follow-up: `AlgorithmState.from_settings` reads `weights.lam_vb_nodal` / `lam_vb_cross` unconditionally, but those default to `None` on raw `Weights(...)` construction. In production `Weights.build_vb_arrays` always runs first, so this never bites — but a guard would make misuse loud.

---


haynec (approved):
LGTM

PR #500, opened 2026-05-16
added better solver args for qpax and moreau

haynec:
Added Solver args

---

Retroactive link: closes #498 — adds the named solver-args API for both JAX-native backends (`QPAXPTRSolver`, `MoreauPTRSolver`) at the constructor and `PTRSolverSpec` levels, with a `solver_args` escape hatch.

Issue #498, opened 2026-05-16
Add api for solver args for Moreau and QPAX


haynec:

griffin-norris:
Resolved retroactively by #500, which added the named solver-args API for both JAX-native backends — keyword-only constructor params (`QPAXPTRSolver(solver_tol=, max_iter=)`; `MoreauPTRSolver(max_iter=, verbose=, device=, tol_gap_abs=, tol_feas=)`), backend-specific `PTRSolverSpec` fields with per-backend validation, and a `solver_args` escape hatch — with #507 as the follow-up fix that made the escape hatch usable (default values no longer trip the overlap check). Residual gap, not blocking: the named params are documented only in docstrings; there's no docs/ page for solver args.

PR #507, opened 2026-05-19
Fix/qpax moreau solver args overlap; pin cvxpy<1.9

griffin-norris:
## Summary
- Fix spurious `ValueError` raised by `QPAXPTRSolver`/`MoreauPTRSolver` whenever a user puts `solver_tol`/`max_iter` (qpax) or `max_iter`/`verbose`/`device` (moreau) inside `solver_args`. The overlap check built `_named` from every kwarg unconditionally, so default values always collided with the same keys in `solver_args` — breaking the documented escape-hatch pattern.
- Switch the contested named-arg defaults to `Optional[...] = None` sentinels; `_named` now only contains keys the caller actually passed. Documented defaults (`solver_tol=1e-5`, `max_iter=30`; `max_iter=200`, `verbose=False`, `device="auto"`) are re-applied via `setdefault` after the overlap check, so the public default contract is unchanged.
- The "user passed both forms" overlap case still raises with the same message — covered by the existing `test_qpax_named_param_overlap_with_solver_args_raises` and its moreau twin.
- Pin `cvxpy<1.9` in `pyproject.toml`. cvxpy 1.9.0 (released 2026-05-18) shifted QOCO failure semantics enough that the unpinned fleet went red overnight on `test_cross_nodal[infeasible-convex]` and a handful of related tests across multiple unrelated branches. Restoring 1.8.x makes CI reproducible while 1.9.x compatibility is investigated separately.

## Test plan
- [x] `pytest tests/test_brachistochrone.py::test_backend_float32_raises` — now passes (was the original failure).
- [x] `pytest tests/solvers/test_qpax_ptr_solver.py` — 15/15 pass, including the explicit-overlap and escape-hatch regression tests.
- [x] `pytest tests/test_brachistochrone.py -m "not integration" -n auto` — 34 passed, 5 unrelated skips.
- [x] CI unit-tests workflow goes green after the cvxpy pin lands.
- [x] Moreau path verified locally by anyone with the `moreau` extra installed (CI still skips moreau tests; symmetric fix follows the same pattern its own unit tests exercise).

---

Retroactive link: part of #498 — fixes the `solver_args` escape hatch from #500, whose overlap check compared default values and spuriously rejected valid overrides.

### DIFF
--- a/.github/workflows/tests-unit.yml
+++ b/.github/workflows/tests-unit.yml
@@ -38,8 +38,11 @@ jobs:
 
     - name: Install dependencies
       run: |
-        # Install package with optional GUI and test packages
-        uv pip install --system -e ".[gui,test]"
+        # Install package with optional GUI, qpax backend, and test packages.
+        # qpax is included so qpax-backed tests don't silently skip via
+        # pytest.importorskip — that's the only thing keeping CI honest
+        # about the qpax solver path.
+        uv pip install --system -e ".[gui,qpax,test]"
 
     - name: Run unit tests
       run: |
--- a/.github/workflows/tests-unit.yml
+++ b/.github/workflows/tests-unit.yml
@@ -38,7 +38,7 @@ jobs:
 
     - name: Install dependencies
       run: |
-        uv pip install --system -e ".[gui,qpax,test]"
+        uv pip install --system -e ".[gui,test]"
 
     - name: Run unit tests
       run: |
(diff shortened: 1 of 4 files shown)


haynec (approved):
LGTM assuming CI passes, I think we can also drop the `cvxpy>=1.8.1`

griffin-norris:
> LGTM assuming CI passes, I think we can also drop the `cvxpy>=1.8.1`

We cannot drop `>=1.81` since that is when the fixes required for sparsity to work went through on their end

PR #512, opened 2026-05-24
Fuse the SCP iteration into one JAX-pure iteration_fn

griffin-norris:
## What this does

Collapses the three-export discretization pipeline plus the Python-side `_subproblem` stitching in `PenalizedTrustRegion` into a single JAX-pure `iteration_fn` (`openscvx/algorithms/scvx/iteration.py`). One call takes `(AlgorithmState, params)` and returns the next state: it discretizes the current iterate (continuous + impulsive), linearizes the nodal/cross-node constraints, packs a `SubproblemData` pytree, hands it to the per-backend solver callback, computes the `J_tr/J_vb/J_vc` convergence metrics, and folds the candidate through the autotuner.

`PenalizedTrustRegion.step()` shrinks from ~80 lines of JIT-warmup/dict-pack scaffolding to a thin wrapper around `iteration_fn`. Also ships `make_solve_loop`, a `lax.while_loop` wrapper over `iteration_fn` (tested; not yet on the `.solve()` path — see Cost).

## How this builds toward a batchable Problem

Foundation slice for `plans/jax-pure-solve.md`. The SCP body is now JAX-pure end-to-end — `state -> state` over a registered pytree — which is the prerequisite for `jax.vmap(problem.solve)` (batch a problem across scenarios) and a `lax.while_loop`-driven `.solve()`. The follow-on plan wraps `iteration_fn` in `make_solve_loop`, registers `OptimizationResults` as a pytree, and adds the batched/grad examples. None of that is in this PR.

To keep the no-disk path vmap-compatible, the discretization compile switches from `jax.export` to plain `jax.jit` (`call_exported` has no vmap batching rule). The disk-cached path (`save_compiled=True`) stays on `jax.export` and is documented as incompatible with `jax.vmap(problem.solve)` until Plan 2 redesigns the cache.

## Cost now / why it isn't breaking

- **No public API change.** `Problem.solve()` signature, return type, dispatch, and behavior are unchanged — same Python `while` loop driving `step()`, same `time_limit`, emitter thread, `continuous` kwarg, return value. Brachistochrone results unchanged within tolerance.
- **Perf gate.** `tests/test_brachistochrone.py` stays within ≤2× baseline on a warm cache (enforced by the per-test timing asserts).
- **`make_solve_loop` is unused in production.** Internal-only primitive (not re-exported), exercised only by `tests/algorithms/test_make_solve_loop.py`. Plan 2 wires it into `.solve()`. Shipping it now lets the `lax.while_loop`-vs-Python-while equivalence be tested in isolation before Plan 2 piles vmap/`OptimizationResults` on top.
- **Error contract shifted (type-preserved).** Solver failures now surface as a `status_code` that `step()` turns into a clean `RuntimeError`, instead of the backend's own guard. The exception *type* is unchanged (`RuntimeError`); the message text and originating layer changed, so two brachistochrone error-path tests were updated to match.

## The CVXPy / QPAX convergence weirdness (worth a look)

Routing the *live* solve through the JAX-pure per-backend callback shook loose two error paths that previously only ran in legacy code:

- **QPAX — `solve_qp` vs `solve_qp_primal`.** qpax exposes two entry points and you can't get both properties at once: `solve_qp_primal` is `custom_vjp`-differentiable but returns *only* the primal (no convergence flag), while `solve_qp` carries a reliable `converged` flag but wraps a `lax.while_loop` that breaks reverse-mode autodiff. This slice needs reliable non-convergence detection (so `step()` can fail loudly), not grad — so it uses `solve_qp` and maps `converged → status_code`. **This reverses the sibling `solver-iteration-callbacks` choice and forfeits differentiability of the QPAX callback** (still `jit`/`vmap`-able). Plan 2 flips the one line back when grad-through-solve is actually on the table. A primal feasibility-residual proxy was tried and rejected — too thin (diverged case violated by only ~5e-5 vs ~1e-16 converged).

- **CVXPy under `pure_callback` — two latent bugs surfaced.** The CVXPy backend hosts its solve via `jax.pure_callback`, and driving the live loop through it exposed that `host_solve` (a) never called `update_proximal_terms()`, so `prox_c` was unset on the first solve of a fresh process — it had only ever been set as a side effect of a prior legacy solve; and (b) let `cvxpy.SolverError` propagate through `pure_callback` as an opaque `ValueError`. Both fixed in `host_solve`: call `update_proximal_terms()` after `update_penalties`, and catch `SolverError → ` non-OPTIMAL `status_code`. Also maps `optimal_inaccurate → OPTIMAL` to preserve the legacy solve tolerance.

One asymmetry worth a reviewer's eye: the two backends now apply *different* leniency before the shared `step()` status gate (CVXPy accepts `optimal_inaccurate`; QPAX maps a hard `converged` bool). Defensible — CVXPy is the only backend with a fuzzy tier — but it's the kind of thing that's easy to trip over later.

## Review notes

Reviewed by a four-lens panel (correctness / minimality / maintainability / API-scope); all must-fixes addressed. Deferred, non-gating should-fixes: a comment naming the QPAX `solve_qp` 6-tuple field order, a note tying the CVXPy `optimal_inaccurate` leniency to the shared status gate, and tightening the `step()` finite-guard target.

---

### DIFF
--- a/docs/UnderTheHood/lowering_architecture.md
+++ b/docs/UnderTheHood/lowering_architecture.md
@@ -168,18 +168,28 @@ of the abstract `PTRSolver` base. Three backends ship today:
 | Backend | Class | Selector | Notes |
 |---------|-------|----------|-------|
 | CVXPy (default) | `CVXPyPTRSolver` | `solver={"backend": "cvxpy"}` (default) | DCP graph via CVXPy, dispatched to QOCO / CLARABEL / etc. Supports user `.convex()` constraints, cross-node constraints, CTCS, and impulsive controls. Optional cvxpygen code generation. |
-| QPAX | `QPAXPTRSolver` | `solver={"backend": "qpax"}` | JAX-native QP via `qpax.solve_qp`. Flat `(Q, q, A, b, G, h)` assembly. Supports box / dynamics (continuous and impulsive) / CTCS / boundary-Fix; **rejects** user `.convex()` and cross-node at `initialize()` with a clear "use `CVXPyPTRSolver`" message. Enables a path toward an end-to-end JAX-differentiable SCP loop in follow-up work. |
+| QPAX | `QPAXPTRSolver` | `solver={"backend": "qpax"}` | JAX-native QP via `qpax.solve_qp`. Flat `(Q, q, A, b, G, h)` assembly. Supports box / dynamics (continuous and impulsive) / CTCS / boundary-Fix; **rejects** user `.convex()` and cross-node at `initialize()` with a clear "use `CVXPyPTRSolver`" message. The `iteration_callback` uses `qpax.solve_qp`, whose convergence flag maps to a `status_code` so a diverged solve fails loudly (chosen over the differentiable `solve_qp_primal`, which exposes no such flag). |
 | Moreau | `MoreauPTRSolver` | `solver={"backend": "moreau"}` | JAX-native conic solver (`moreau.jax.Solver`). Sparse CSR assembly; SOC epigraphs for the L1 / pos PTR penalties (fewer variables and rows than QPAX). Warm-starts between SCP iterations. Same supported subset as QPAX (continuous and impulsive dynamics). Paves the way for user `.convex()` SOC support in a follow-up. |
 
 Each backend exposes two per-iteration entry points: the historical NumPy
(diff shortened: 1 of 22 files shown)


haynec (approved):
LGTM

PR #514, opened 2026-05-28
JAX-pure Problem.solve_jax() alongside Problem.solve()

### LINKED ISSUE #493: Make Entire Solve Call Batchable

griffin-norris:
## Summary

- Adds `Problem.solve_jax()` — a JAX-pure entry point composable with `jax.vmap` and `jax.jit` — alongside the unchanged `Problem.solve()`. Two methods, zero dispatch: users pick by name based on whether they need batching/JIT or prints/`time_limit`/history.
- `OptimizationResults` registers as a JAX pytree; new `from_final_state` (JAX path) / `from_history` (Python path) constructors split the empty-history vs full-history asymmetry honestly.
- `make_solve_loop`'s `lax.while_loop` body pins already-converged batch elements to their first post-convergence iterate, so `jax.vmap(problem.solve_jax)` agrees with per-element `solve_jax` (no drift from extra iterations after convergence).
- `_solve_loop_fn` warms **lazily** on first `solve_jax()` call rather than eagerly in `initialize()`. See the benchmark + rationale below.
- Moreau cold-start framing corrected: post-Plan-1 it's cold-start on **both** paths today; only the legacy NumPy `solve()` entry point (no longer on the SCP hot path) warm-starts. Docstrings updated to say so honestly.
- Four new tests (bare / vmap / jit / no-drift); new `examples/abstract/brachistochrone_batched.py` and `docs/UnderTheHood/batching_jit_grad.md`.

Not in scope: changes to `Problem.solve()` (kept verbatim), validated `jax.grad` through `solve_jax`, Moreau warm-start under the JAX path, `lax.scan`-based per-iteration history.

## Why warm `solve_jax` lazily instead of in `initialize()`

The first cut of this PR pre-warmed `_solve_loop_fn` (the `jax.jit`'d `lax.while_loop` backing `solve_jax`) at the end of `Problem.initialize()`, so the user's first `solve_jax` call would land on a populated cache. That choice had two problems:

1. **`.solve()`-only users paid for work they never used.** `_iteration_fn` warmup at the same site helps both paths (the Python loop driving `.solve()` calls into it too), but `_solve_loop_fn` is exclusive to `solve_jax`. Every problem — including the brachistochrone test suite, every example, every `.solve()`-only user — was paying ~2.5s of XLA compile work for a code path they would never invoke.
2. **It broke Ubuntu CI.** The brachistochrone tests assert `timing_init < 15s`. macOS CI absorbed the extra ~2.5s; Ubuntu CI's slower XLA compile pushed `timing_init` to 15.2–17.3s across all 4 Python versions (3.11–3.14), failing every Ubuntu job.

We evaluated three fixes:
- **Bump the threshold to 20s.** Cheapest, but hides a real regression and the cost stays on every user.
- **Move the warmup outside the timed window.** Fixes CI but `.solve()` users still pay.
- **Warm lazily on first `solve_jax` call (chosen).** First `solve_jax` pays the XLA trace; subsequent calls hit `jax.jit`'s shape-keyed cache, identical to the eager-warm world. `.solve()`-only users pay nothing. Total compile work is unchanged — it just lives at the call site that asked for it. MPC users wanting a primed cache before a latency-critical inner loop can run a throwaway `problem.solve_jax()` themselves (standard JAX warm-up idiom, called out in the `_get_or_build_solve_loop` docstring).

### Benchmark (QPAX, N=8, brachistochrone, median of 3)

| | Eager warmup in `initialize()` | Lazy warmup on first `solve_jax()` |
|---|---|---|
| `initialize()` | 4.71s | **2.17s** (−54%) |
| First `solve_jax()` | 0.68s | 2.44s |
| Second `solve_jax()` (cached) | 0.68s | 0.67s |
| `.solve()` (Python loop) | 0.71s | 0.71s |

Total init + first-`solve_jax` is ~4.6s in both worlds — the same XLA trace, just attributed to the call site that needs it.

## Test plan

- [x] `pytest -n auto -m 'not integration' tests/algorithms/ tests/test_brachistochrone.py tests/test_optimization_results.py tests/test_solve_jax_*.py` green (locally: 65 passed, 5 skipped).
- [x] Confirmed Ubuntu `timing_init` regression resolved by lazy-warm (brachistochrone now ~2.2s, well under 15s threshold).
- [ ] Spot-check `examples/abstract/brachistochrone_batched.py` runs end-to-end and prints batched shapes / `t_final` / `converged` for the four ICs.
- [ ] Confirm `jax.vmap(problem.solve_jax)` on the CVXPy backend runs B sequential CVXPy solves (no host-CVXPy thread-safety violation).

## Follow-ups (not in this PR)

- Inline two single-call-site `Problem` helpers (`_resolve_initial_state`, `_format_result`) per panel review.
- Fix carried-forward `from_history` bug writing `float(...)` into `list[np.ndarray]`-typed `J_*_history` fields.
- Sketch landed at `plans/solve-batched-export.md` for the eventual exportable, disk-cached `solve_batched`.

---

Retroactive link: closes #493 — with `Problem.solve_jax()` the entire solve call composes with `jax.vmap`/`jax.jit`. Part of the batchable-solve line for #494.

### DIFF
--- a/docs/UnderTheHood/batching_jit_grad.md
+++ b/docs/UnderTheHood/batching_jit_grad.md
@@ -4,9 +4,9 @@
 Python-loop driver — real-time prints, wall-clock `time_limit`, `continuous`
 mode, populated per-iteration history. `solve_jax()` is its JAX-pure
 sibling: it drives the same fused `iteration_fn` body inside a
-`lax.while_loop` and returns an
-[`OptimizationResults`](../../openscvx/algorithms/optimization_results.py)
-pytree, so it composes with `jax.vmap`, `jax.jit`, and `jax.grad`.
+`lax.while_loop` and returns an `OptimizationResults`
+(`openscvx/algorithms/optimization_results.py`) pytree, so it composes with
+`jax.vmap`, `jax.jit`, and `jax.grad`.
 
 ```python
 problem = ox.Problem(..., solver={"backend": "qpax"})
@@ -27,8 +27,7 @@ batched = jax.vmap(problem.solve_jax, in_axes=(0, 0, None))(
 fast_solve = jax.jit(problem.solve_jax)
 ```
 
-A worked example lives at
-[`examples/abstract/brachistochrone_batched.py`](../../examples/abstract/brachistochrone_batched.py)
+A worked example lives at `examples/abstract/brachistochrone_batched.py`
 — four brachistochrone problems with shifted starting x-coordinates,
 solved in parallel via `jax.vmap(problem.solve_jax)`.
 
@@ -52,9 +51,8 @@ discussion.
 
 ## What `solve_jax()` returns
 
-`solve_jax()` returns the same
-[`OptimizationResults`](../../openscvx/algorithms/optimization_results.py)
-type as `solve()`, but built via
+`solve_jax()` returns the same `OptimizationResults` type as `solve()`,
+but built via
 `OptimizationResults.from_final_state(state, problem=...)` instead of
 `from_history(history, final_state, ...)`. The differences:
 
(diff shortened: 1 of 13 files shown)


haynec (approved):
LGTM

PR #525, opened 2026-06-10
solve_batched: unified rank-based batching API with in_axes override

griffin-norris:
Follow-up to #520. `solve_batched` now shares `solve_jax`'s vocabulary with one rule on top: **add a leading batch axis to whatever varies across solves.**

## The new syntax

Same convention as `jax.vmap`: a batched input carries one extra leading axis (`in_axes=0`), a shared input keeps its natural shape (`in_axes=None`). `solve_batched` infers that axis spec by rank against declared shapes and applies the vmap internally — results gain a leading `B` axis, like vmap outputs.

```python
B = 4  # batch size. Just an int for building the arrays below — there is no
       # B argument; solve_batched infers it from leading-axis lengths (like jax.vmap)
results = problem.solve_batched(
    x_initial=x0_batch,                          # (B, n_x) -> batched boundary pins
    parameters={"gate_center": centers,          # (B, 3) vs declared (3,) -> batched
                "gate_radius": 0.5},             # () -> shared
    algorithm={                                  # same names/box as the constructor's
        "ep_tr":    jnp.logspace(-5, -3, B),     #   (B,) vs field () -> tolerance sweep
        "lam_prox": jnp.array([1.0, 2.0, 4.0]),  #   (B,) fill -> one weight per element
    },
    max_iters=jnp.array([5, 10, 20]),            # sugar for algorithm={"k_max": ...}
    # in_axes={"parameters": 0}                  # optional explicit override — vmap's
    #                                              # vocabulary incl. prefix semantics
)
```

There is no whitelist behind `algorithm`: it takes the same names (and box) as the `Problem` constructor's `algorithm` dict, derived from the `AlgorithmState` fields plus whatever the autotuner declares — solve-time mirrors construction-time. Writing a new autotuner gets batching for free — declaring the knob *is* the registration:

```python
class MyHyper(HyperParams):          # no decorator, no pytree code —
    ramp: float = 2.0                # the annotation drives the dtype

class MyAutotuner(AutotuningBase):
    def __init__(self, ramp=2.0):
        self.hyper = MyHyper(ramp=ramp)
    def update_weights(self, state, ...):
        ... state.hyper.ramp ...     # typed attribute read at trace time

problem.solve_batched(algorithm={"ramp": jnp.linspace(1, 4, B)})  # just works
```

## Changes

- **Unified signature** — `solve_batched` takes `solve_jax`'s argument names; `x0_stack`/`xf_stack` → `x_initial`/`x_final` (breaking, no shim — the method shipped days ago in #520).
- **Rank-based inference** — an input is batched iff it has exactly one more axis than its declared shape, never guessed from leading-axis values. Fixes two misclassifications in the old inference (a lone shared vector faking `B`; a shared matrix with leading axis `== B` being wrongly sliced).
- **`in_axes` override** — explicit escape hatch with `jax.vmap`'s vocabulary *and prefix semantics*: a bare `0` batches every passed input, `in_axes={"parameters": 0}` batches a whole subtree, per-key dicts pick out entries; partial specs fall back to the rank rule. (One documented divergence: the default infers by rank instead of vmap's batch-everything.)
- **Solve-time mirrors construction-time: the `algorithm={...}` dict** — every SCP constant (`ep_*`, `k_max`, weights `lam_prox`/`lam_vc`/`lam_cost`, autotuner knobs) is a runtime input on the `AlgorithmState` pytree, reachable on both solve methods through the same names and box the constructor uses. No whitelist (the name set derives from state fields + declared hyperparameters); array-shaped fields accept scalar/`(B,)` fill forms so weight sweeps don't require full stacks; construction-only keys (`autotuner`, `t_max`) get a teaching error pointing back at the constructor. Components are the namespace — a future `discretizer=`/`solver=` kwarg is the same mechanism with another box.
- **`HyperParams` container** — subclassing handles the frozen-dataclass transform, pytree + `jax.export` registration, and annotation-driven dtypes; flat `__init__` kwargs still work, preserving the YAML loader path. Proven by a test defining an autotuner entirely in-test and sweeping its knob with zero library edits.
- **Cache keys inverted accordingly** — tolerances, `max_iters`, weights, and hyperparameters no longer retrace or re-export (one `save_compiled` artifact serves every setting); the shared/batched parameter split now correctly keys both the in-memory closure and the disk-export hash.
- **Single convergence test** — `_converged(state)` shared by the pure loop, interactive `step()`, and result formatting (which previously read `algorithm.ep_*` and would misreport per-element convergence under sweeps).
- **Teaching errors** — unknown parameter/override names (listing the valid ones), shape mismatches (naming every accepted form), disagreeing batch sizes, kwarg/override collisions, and nothing-batched all raise before tracing; the docstring lists what can never be batched (`N`, backend, discretizer, autotuner type, dtype) and why.
- **Tests, examples, docs** — resolver unit tests incl. fill forms; element-wise sweep equivalence vs per-element `solve_jax` for tolerances, iteration budgets, weights, and a user-defined autotuner knob; export-roundtrip asserts artifact reuse across `max_iters` and re-keying across parameter splits; examples/docs migrated (also fixes `brachistochrone_batched.py`, whose `solve_batched` call passed positionals the shipped signature rejected).

## Notes for review

- `AugmentedLagrangian.rho_init`/`rho_max` turned out to have **no read sites** (the ρ in `update_weights` is the per-iteration acceptance ratio, governed by `eta_*`). They're declared on the AL `HyperParams` so the names resolve, but sweeping them is a no-op — wire up or delete in a follow-up.
- Component-level numerics (discretizer `rtol`/`atol`, QPAX `solver_args`) are **not** in the override surface — they're baked into separately exported artifacts. The `HyperParams` container is reusable for that promotion; deferred.
- Not verified locally: the two moreau-gated drone examples (API-key extra). Call sites unchanged; the same path is covered by QPAX parameter-batching tests.

---

### DIFF
--- a/openscvx/algorithms/autotuner/adaptive_proximal_weight.py
+++ b/openscvx/algorithms/autotuner/adaptive_proximal_weight.py
@@ -13,13 +13,19 @@ from pydantic import BaseModel, ConfigDict
 
 from openscvx.config import Config
 
-from ..base import AdaptiveStateCode, AutotuningBase
+from ..base import AdaptiveStateCode, AutotuningBase, HyperParams
 from .augmented_lagrangian import AugmentedLagrangian
 
 if TYPE_CHECKING:
     from openscvx.lowered import LoweredJaxConstraints
 
     from ..base import AlgorithmState, CandidateIterate
+
+
+class AdaptiveProximalWeightHyper(HyperParams):
+    """Declared hyperparameters for :class:`AdaptiveProximalWeight`."""
+
+    lam_cost_drop: int = -1
 
 
 class AdaptiveProximalWeight(AutotuningBase):
@@ -54,7 +60,7 @@ class AdaptiveProximalWeight(AutotuningBase):
         self.eta_2 = eta_2
         self.lam_prox_min = lam_prox_min
         self.lam_prox_max = lam_prox_max
-        self.lam_cost_drop = lam_cost_drop
+        self.hyper = AdaptiveProximalWeightHyper(lam_cost_drop=lam_cost_drop)
         self.lam_cost_relax = lam_cost_relax
 
     def update_weights(
@@ -85,7 +91,7 @@ class AdaptiveProximalWeight(AutotuningBase):
         J_nonlin = nonlin_cost + nonlin_pen + nodal_pen
 
         lam_cost_next = jnp.where(
-            state.k > state.lam_cost_drop,
+            state.k > state.hyper.lam_cost_drop,
             state.lam_cost * self.lam_cost_relax,
             state.lam_cost_init,
         )
--- a/openscvx/algorithms/autotuner/adaptive_proximal_weight.py
+++ b/openscvx/algorithms/autotuner/adaptive_proximal_weight.py
@@ -85,7 +85,7 @@ class AdaptiveProximalWeight(AutotuningBase):
         J_nonlin = nonlin_cost + nonlin_pen + nodal_pen
 
         lam_cost_next = jnp.where(
-            state.k > self.lam_cost_drop,
+            state.k > state.lam_cost_drop,
             state.lam_cost * self.lam_cost_relax,
             state.lam_cost_init,
         )
--- a/openscvx/algorithms/autotuner/augmented_lagrangian.py
+++ b/openscvx/algorithms/autotuner/augmented_lagrangian.py
@@ -244,12 +244,14 @@ class AugmentedLagrangian(AutotuningBase):
         )
         J_nonlin = nonlin_cost + nonlin_pen + nodal_pen
 
-        # Cost relaxation: when state.k > lam_cost_drop, scale state.lam_cost;
-        # otherwise reset to the algorithm's initial weight (carried on the
-        # pytree as state.lam_cost_init, broadcast at from_settings()). Scalar
+        # Cost relaxation: when state.k > state.lam_cost_drop, scale
+        # state.lam_cost; otherwise reset to the algorithm's initial weight
+        # (carried on the pytree as state.lam_cost_init, broadcast at
+        # from_settings()). Both constants ride the pytree so per-solve
+        # overrides and vmap sweeps reach the traced body. Scalar
         # lam_cost_relax preserves the user-specified per-state weight ratios.
         lam_cost_next = jnp.where(
-            state.k > self.lam_cost_drop,
+            state.k > state.lam_cost_drop,
             state.lam_cost * self.lam_cost_relax,
             state.lam_cost_init,
         )
--- a/openscvx/algorithms/autotuner/augmented_lagrangian.py
+++ b/openscvx/algorithms/autotuner/augmented_lagrangian.py
@@ -19,12 +19,20 @@ from openscvx.utils.printing import (
     color_adaptive_state,
 )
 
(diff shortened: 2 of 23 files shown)


openscvx/problem.py
```diff
@@ -1474,60 +1530,61 @@ def solve_jax(
...
-
-        **Batching over parameters**: declare any state or constraint whose
-        value changes across solves as an ``ox.Parameter``, then pass an
-        ``(B, ...)``-shaped array for it in *parameters*::
-
-            ic_param = ox.Parameter("ic", shape=(3,), value=default_ic)
-            constraints += [(state == ic_param).convex().at([0])]
-            ...
-            results = problem.solve_batched(parameters={"ic": ic_batch})
+        """Run ``B`` SCP solves in parallel as one batched XLA program.
```
haynec, line 1754: This seems a little arbitrary and not general to manually take in these specific parameters and not others, say lam_prox for example. I was hoping we could have a unified way of batching over any of these parameters without the need to have a bunch of arguments for this function

haynec (changes requested):

haynec (approved):
LGTM

openscvx/algorithms/autotuner/augmented_lagrangian.py
```diff
@@ -19,14 +19,22 @@
...
 
     from ..base import AlgorithmState, CandidateIterate
 
 
+class AugmentedLagrangianHyper(HyperParams):
+    """Declared hyperparameters for :class:`AugmentedLagrangian`."""
+
+    rho_init: float = 1.0
+    rho_max: float = 1e2
+    lam_cost_drop: int = -1
```
haynec, line 35: Can we not have more included hyper parameters here? Ideally all of them

```        
rho_init: float = 1.0,
rho_max: float = 1e2,
gamma_1: float = 2.0,
gamma_2: float = 0.5,
eta_0: float = 1e-2,
eta_1: float = 1e-1,
eta_2: float = 0.8,
ep: float = 0.99,
eta_lambda: float = 1e1,
lam_vc_max: float = 1e5,
lam_prox_min: float = 1e-3,
lam_prox_max: float = 1e4,
lam_cost_drop: int = -1,
lam_cost_relax: float = 1.0,```

File: openscvx/algorithms/autotuner/adaptive_proximal_weight.py [hunk 3]
```diff
@@ -22,6 +22,12 @@
     from ..base import AlgorithmState, CandidateIterate
 
 
+class AdaptiveProximalWeightHyper(HyperParams):
+    """Declared hyperparameters for :class:`AdaptiveProximalWeight`."""
+
+    lam_cost_drop: int = -1
```
haynec, line 28: Again can we have all the parameters for this tuning method included?

openscvx/algorithms/autotuner/ramp_proximal_weight.py
```diff
@@ -11,14 +11,20 @@
...
 if TYPE_CHECKING:
     from openscvx.lowered import LoweredJaxConstraints
 
     from ..base import AlgorithmState, CandidateIterate
 
 
+class RampProximalWeightHyper(HyperParams):
+    """Declared hyperparameters for :class:`RampProximalWeight`."""
+
+    lam_cost_drop: int = -1
```
haynec, line 25: Same as the others.

PR #527, opened 2026-06-10
Fix double-applied node-0 impulse in post-processing

spadafab:
## Problem

For impulsive problems the post-processed trajectory applies the first burn twice (starts from `x.initial + 2·dv[0]` instead of `x.initial + dv[0]`). `x[0]` is the post-impulse node-0 state, but post-processing seeds the re-propagation from it and then re-applies `dynamics_discrete` at node 0.

## Regression source

Introduced in #520 (`6b1cd407`), which replaced the impulse-safe seed `jnp.where(mask, x[0,:], x_prop.initial)` with an unconditional `x[0,:]`. The old `where` had been feeding the pre-impulse value for fixed initial states; collapsing it dropped that protection.

## Fix

Seed node 0 from the pre-impulse state when impulses are present, mirroring the solver's node-0 prior (`where(init_fixed, x.initial, x[0])`). Only activates for impulsive problems, so #520's non-impulsive behavior is unchanged.

## Verification

Verified on the Hohmann test (start velocity corrected to `v_initial + dv[0]`; `tests/test_impulsive.py` passes). Contained to `openscvx/propagation/post_processing.py`.

### DIFF
--- a/openscvx/propagation/post_processing.py
+++ b/openscvx/propagation/post_processing.py
@@ -81,18 +81,23 @@ def propagate_trajectory_results(
     n_opt_states = x.shape[1]
     n_prop_states = settings.sim.x_prop.initial.shape[0]
 
-    # For a converged solution x[0, :] is the actual IC for every optimisation-state
-    # component regardless of initial_type (Fix / Free / Maximize).  Using the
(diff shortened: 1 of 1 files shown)


griffin-norris (approved):
Lgtm

PR #529, opened 2026-06-11
Restore the Algorithm / Autotuner contracts

griffin-norris:
Follow-up to #525. Restores the base-class contracts that drifted during the JAX-purity rewrite, so a new dev can drop in a new `Algorithm` or autotuner against the documented ABCs alone — and resolves @haynec's #525 review comments asking that **all** autotuner parameters be declared hyperparameters.

## What changed

- **Every constructor knob of every autotuner is now a declared hyperparameter** (`gamma_1/2`, `eta_0/1/2`, `ep`, `eta_lambda`, the weight clips, `ramp_factor`, the cost-relaxation schedule, …). Declaration is registration: each knob is a per-solve override and a batched sweep target by name, with no core edits. Constructor signatures are unchanged, and knobs remain readable/writable as plain attributes via a proxy that routes into the frozen `hyper` container.

  ```python
  # the #525 ask, end-to-end: sweep a trust-region growth factor per batch element
  results = problem.solve_batched(algorithm={"gamma_1": jnp.linspace(1.5, 3.0, 16)})
  ```

- **`Algorithm` owns its iteration again.** `Problem` no longer constructs the PTR-specific fused body; it hands the compiled ingredients to the new abstract `build_iteration` and stays algorithm-agnostic. The ABC now declares the full surface `Problem` touches (`autotuner`, `ep_*`, `get_columns`, a default `_hash_into`) — a subclass written from the docstrings alone runs through `initialize()`/`solve()` (verified empirically during review).
- **~120-line duplicated `update_weights` collapsed** into `AcceptanceRatioAutotuner` with a single `_update_multipliers` hook; `AugmentedLagrangian` supplies the constraint-violation rule, `AdaptiveProximalWeight` is the hook-default. Verified behavior-identical to `main` line-by-line; the `solve_batched` artifact for default settings hashes to the same key across the `build_iteration` move.
- **Hardcoded autotuner roster removed from `step()`** — replaced by a declared `COMPUTES_ACCEPTANCE_METRICS` capability flag, so custom tuners get correct diagnostics handling.
- **Dead migration API deleted**: the unread `JIT_UPDATE_WEIGHTS` flag, `AlgorithmHistory`'s never-called discretization accessors, `CandidateIterate`'s never-set fields, and the hand-maintained `AlgorithmState._FIELDS` tuple (now derived). Stale docstrings corrected (`lam_cost_drop=-1` semantics, REJECT recording, the dead `Dis (ms)` column).
- **Warning hygiene**: removed three blanket process-wide `warnings.filterwarnings("ignore")` calls (PTR module + two in `utils/printing.py`) that silenced every warning for users; the known CVXPY noise is re-silenced with message-scoped filters in the cvxpy backend only. `rho_init`/`rho_max` stay as reserved (currently unused) knobs and now warn on non-default values.
- **Tests**: new coverage pins all three knob channels per autotuner, the capability flag, `build_iteration` parity with the bare factory, the reserved-knob warning, and a batched `gamma_1` sweep matching per-element `solve_jax`. Docs (`batching_jit_grad.md`) refreshed.

Custom autotuners follow the same recipe — the declaration is the registration:

```python
class MyHyper(HyperParams):
    ramp: float = 2.0

class MyAutotuner(AutotuningBase):
    def __init__(self, ramp: float = 2.0):
        self.hyper = MyHyper(ramp=ramp)

    def update_weights(self, state, candidate, *args):
        return state.replace(
            x=candidate.x, u=candidate.u,                 # accepted candidate -> next iterate
            x_prop=candidate.x_prop, x_prop_plus=candidate.x_prop_plus,
            lam_prox=state.lam_prox * state.hyper.ramp,   # knob rides state.hyper
            adaptive_state_code=...,
        )
```

## Notes for reviewers

- One-time `solve_batched` cache-key change: promoted knobs left the export hash (they are runtime inputs now), so existing exported artifacts rebuild once. One artifact then serves every knob setting.
- Two qpax-gated tests fail identically on `main` and this branch (`test_solve_batched_matches_solve_jax_over_x_initial[qpax]`, `test_solve_batched_ep_tr_sweep_matches_solve_jax`) — pre-existing, not introduced here; possibly qpax-version sensitivity worth a separate issue.
- The CVXPY `*`-as-matmul deprecation in the subproblem assembly is knowingly filtered, with the `*`→`@` fix tracked as follow-up work.

---


haynec (approved):
LGTM

PR #533, opened 2026-06-17
Chore/restore dis solve timing

haynec:
**Summary**
Restore separate per-iteration timing for discretization and convex solve on the Python solve() path. The fused SCP iteration body is split into make_scp_prepare and make_scp_finalize, and PenalizedTrustRegion.step() reports Dis (ms) and Cvx (ms) instead of a single Step (ms) column.


plans/scp-split-iteration-timing.md
griffin-norris, line 1: I thought we weren't committing plans?

griffin-norris (commented):
LGTM except the committed plan file

PR #535, opened 2026-06-18
Refactor/better multishot interface

haynec:
Summary
Add a library-level API for reading SCP multishot propagated trajectories from discretization_history, replacing duplicated V-matrix unpackers across examples and plotting code.

- Introduce MultishotPropagation and `unpack_multishot_V()` in `openscvx/algorithms/multishot.py` with `segments()`, `chronological()`, and `symbolic .state(name | State) `access
- Add OptimizationResults.multishot_propagation(iteration=-1) as the primary entry point; returns None when history is empty (e.g. `solve_jax`)
- Raise ValueError on t_nodes length mismatch instead of silently repairing or failing
- Refactor `extract_propagation_positions`, `plot_scp_iterations`, and `extract_multishoot_trajectory` to use the shared unpacker; chronological stitching replaces the old substep-major ordering in realtime helpers
- Migrate frax, mjx, and flappy bird examples to the new interface


openscvx/algorithms/multishot.py
```diff
@@ -0,0 +1,158 @@
+"""Read-side helpers for SCP multishot integration matrices ``V``.
+
+:class:`MultishotPropagation` complements
```
spadafab: This contribution would fit better in the file `history.py` instead of a completely new file

spadafab (commented):
looks good. Prior to merging, the` multishot.py` contents should be factored inside the script `history.py` as suggested in the comments

openscvx/algorithms/multishot.py
haynec: Addressed

spadafab (dismissed):
ready for me

spadafab (approved):
Issues in comments addressed

PR #544, opened 2026-06-30
Support nonconvex equality constraints as L1-penalized nodal/cross-node constraints

spadafab:
## Support nonconvex equality constraints as L1-penalized nodal/cross-node constraints

Previously, OpenSCvx rejected any nonconvex equality constraint at the symbolic
augmentation stage with an explicit `ValueError`, requiring users to manually
reformulate non-convex equalities as pairs of inequalities or restrict themselves
to convex (affine) equalities only.

With this PR, non-convex equalities are fully supported as penalty constraints.

- **Removed the hard block** in `augmentation.py` — `_check_nonconvex_equality`
  and all four call sites are gone; non-convex equalities now flow through the
  same lowering path as inequalities.
- **`is_equality` flag propagated through lowering** — `LoweredNodalConstraint`
  and `LoweredCrossNodeConstraint` each carry a new `is_equality: bool` field,
  set during JAX lowering by inspecting whether the source constraint is an
  `Equality` instance.
- **Two-sided L1 penalty for equalities** — wherever virtual-buffer violations
  were previously computed with `max(0, g)` (one-sided, for inequalities), they
  now branch on `is_equality`: equalities use `|g|` (two-sided), inequalities
  keep `max(0, g)`. This applies in both `penalty.py` (nonlinear penalty
  evaluation) and `iteration.py` (linearized `J_vb` inside the SCP loop).

### Effect

Users can now directly express non-convex equality constraints
(e.g. `f(x, u) == 0` where `f` is nonlinear) and have them handled correctly
by the L1 penalty machinery, without any manual reformulation.

### DIFF
--- a/openscvx/algorithms/autotuner/augmented_lagrangian.py
+++ b/openscvx/algorithms/autotuner/augmented_lagrangian.py
@@ -206,7 +206,7 @@ class AugmentedLagrangian(AcceptanceRatioAutotuner):
 
         for idx, constraint in enumerate(nodal_constraints.nodal):
             g = constraint.func(candidate.x, candidate.u, 0, params)
-            nu = jnp.maximum(0.0, g)
+            nu = jnp.abs(g) if constraint.is_equality else jnp.maximum(0.0, g)
 
             if constraint.nodes is not None:
                 nodes_array = jnp.asarray(constraint.nodes)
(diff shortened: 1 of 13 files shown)


haynec (approved):
LGTM

PR #531, opened 2026-06-15
Chore/aircraft sstc example

griffin-norris:
Adds the classic Bryson / GPOPS-II **minimum time-to-climb of a supersonic aircraft** benchmark as an OpenSCvx example. -> Fixes #207

- **Problem.** Free-final-time, point-mass dynamics over a spherical Earth (states `[h, v, γ, m]`, control α); minimize time to climb from sea level to 20 km. Reaches **t_f ≈ 324.9 s** — matching GPOPS-II's 324 s and propagation-consistent (the residual is the bilinear-thrust model gap).
- **Tabulated data & interpolation.** Cubic `Cinterp` for the smooth atmosphere tables, **PCHIP** `Cinterp` for the aero coefficients (shape-preserving — avoids a cubic's non-physical overshoot at the sharp transonic peak), and `Bilerp` for the 2-D engine-thrust table.
- **Energy-maneuverability solution.** A one-line "hold near the altitude floor, then climb" initial guess steers the local SCP into the global *sea-level dash → zoom-climb* basin rather than the suboptimal *climb-immediately* local minimum.
- **Solver config.** `ConstantProximalWeight` (`lam_prox=2e-2`) plus a tightened `licq_max` converge cleanly and fast (CTCS violation ~3e-11, zero floor sag).
- **Diagnostics.** `plot_gpops_comparison()` reproduces the four GPOPS Figure-8 plots (altitude–time, altitude–speed energy path, flight-path-angle–time, angle-of-attack–time) with the propagated trajectory and nodes overlaid.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #546, opened 2026-07-05
Chore/aircraft dynamic soaring example

### LINKED ISSUE #208: GPOPS-II Example: Dynamic Soaring Problem

griffin-norris:
## Summary

Adds the Zhao 2004 / GPOPS-II dynamic soaring benchmark as `examples/aircraft/dynamic_soaring.py` -> Fixes #208, companion to the supersonic time-to-climb example (#531). A glider flies one closed loop through a linear wind shear; the objective is the smallest wind-gradient slope `beta` for which an energy-neutral loop exists.

Formulation mapping from the GPOPS problem:
- **Free static parameter `beta`** → a zero-dynamics state whose final value carries a `Minimize` boundary condition (the Mayer objective).
- **Periodicity events** (`v_f = v_0`, `gamma_f = gamma_0`, `psi_f = psi_0 - 2pi`) → affine cross-node equality constraints marked `.convex()`, enforced exactly by the subproblem solver.
- **Load-factor path constraint** (`n ∈ [-2, 5]`) and state boxes → CTCS constraints, satisfied in continuous time between nodes — 20 nodes suffice where the reference refines an adaptive collocation mesh.

`plot_gpops_comparison` reproduces the guide's six diagnostic panels: 3D x-y-h soaring loop, then airspeed, flight-path angle, heading, lift coefficient, and bank angle vs. time (propagated line + node markers).

## Results

- Converges in ~25 SCP iterations (fixed proximal weight), solve time ~0.3 s at N=20.
- `beta* ≈ 0.069 1/s`, loop period ≈ 20.3 s; the solution rides the n = 5 load-factor ceiling through the high-speed bottom turn.
- Forward propagation closes the loop to ~1e-3 ft and is periodic in v/gamma/psi to ~1e-5 — the loop is feasible in continuous time.
- The problem is nonconvex; this is a slightly different local basin than the IPOPT solution pictured in the guide (which shows v spanning ~45-240 ft/s vs. our 73-211). Noted in the module docstring.

## Test plan

- [x] `python examples/aircraft/dynamic_soaring.py` converges and renders all four figures
- [x] Solution validated by forward propagation (loop closure, periodicity, load factor within limits between nodes)
- [x] `beta*` consistent across N = 20/25/30/50/60 (~0.5% spread)
- [x] ruff check / format clean

### DIFF
--- a/examples/aircraft/dynamic_soaring.py
+++ b/examples/aircraft/dynamic_soaring.py
@@ -1,0 +1,333 @@
+"""Minimum wind-gradient dynamic soaring of a glider (Zhao 2004).
+
+The dynamic soaring benchmark from Zhao, *"Optimal Pattern of Glider Dynamic
+Soaring,"* Optimal Control Applications and Methods, Vol. 25, 2004, pp. 67-89,
+reproduced in the GPOPS-II user's guide. A glider flies one closed loop through
+a linear wind shear, extracting enough energy from the gradient to return to
+its initial airspeed, flight-path angle, and (heading - 2 pi). The objective is
+the *smallest* wind-gradient slope ``beta`` for which such an energy-neutral
+loop exists.
+
+Point-mass dynamics over a flat Earth with horizontal wind ``W_x = beta h``
+(state ``[x, y, h, v, gamma, psi]`` = position, airspeed, flight-path angle,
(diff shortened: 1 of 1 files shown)


spadafab (approved):
Looks good

PR #547, opened 2026-07-05
Fix/results parameters

griffin-norris:
## Problem

Post-processing propagates the converged solution with the Problem's *current* parameter dict (`self._parameters`), not the values the solve actually used. The dense trajectory is silently corrupted whenever the two differ:

- **`solve_batched(parameters=...)` with per-element values that enter the dynamics** — every element was propagated with the shared `self._parameters` (this limitation was documented on `post_process_batched`; this PR removes it).
- Mutating `problem.parameters` between `solve()` and `post_process()`.

Found while batch-solving a hybrid race car example over powertrain parameters (`mgu_k`, `ice_share`): all three cars' dense trajectories came back propagated as the hybrid.

## Fix

Results now record what happened:

- `solve()` and `solve_batched()` snapshot the merged, as-used parameter values onto `results.parameters` (plain numpy copies). Batched solves store every value **post-broadcast** — each leaf carries the leading `(B,)` axis, shared values replicated, explicit non-zero `in_axes` moved to the front — reusing the existing rank-rule resolution.
- `post_process()` / `post_process_batched()` propagate with the recorded values; `post_process_batched` indexes element `b`'s own values in its existing per-element loop. Only the batched path keeps a fallback to `self._parameters` — its `results` argument may be loaded from a save that predates the snapshot — and it emits a `UserWarning` when taken, since the propagation values may then differ from the solved ones. `post_process()` needs no fallback: `solve()` is the only setter of the solution it reads and always records the snapshot with it.
- The rule, documented on both methods: **results record what happened; the Problem's parameter dict only seeds the next solve.**

The snapshot rides the `OptimizationResults` treedef's **aux** alongside `plotting_data` — pytree children are unchanged, so `solve_jax`'s `vmap`/`jit` composability is untouched.

> [!NOTE]
> **`solve_jax` deliberately does not snapshot.** Two reasons, documented on its `parameters` arg:
> 1. Under a caller's `jit`/`vmap` the merged parameter values are **tracers**. The snapshot lives in the pytree's aux, which must hold static data — attaching tracers would leak them, and attaching only-when-untraced would make the output treedef differ between traced and untraced calls.
> 2. It is moot for the built-in pipeline: `solve_jax` never sets `self._solution`, so `problem.post_process()` raises on its output regardless — there is no library propagation path downstream of `solve_jax` for a snapshot to protect. Its callers thread parameters through their own code and do their own propagation bookkeeping.
>
> `solve()` and `solve_batched()` are the two entry points that feed the library's own post-processing, and both record.

## Tests

`tests/test_results_parameters.py` (fixture: brachistochrone whose `gain` parameter scales the position kinematics, so a wrong propagation parameter is directly visible):

- solve → mutate `problem.parameters` → post_process still propagates with the solved values
- **headline:** `solve_batched` with per-element gains `[0, 1]` → element 0's dense trajectory stays frozen, element 1 moves; snapshot shape assertions
- results without a snapshot fall back to `self._parameters`

All three end-to-end tests verified to **fail on the unfixed source** and pass with the fix. Plus save/load round-trip and pytree-invariance tests in `tests/test_optimization_results.py`; existing solve_batched / solve_jax / propagation suites pass unchanged.

### DIFF
--- a/openscvx/algorithms/optimization_results.py
+++ b/openscvx/algorithms/optimization_results.py
@@ -50,6 +50,14 @@ class OptimizationResults:
             Added by propagate_trajectory_results.
         ctcs_violation (Optional[np.ndarray]): Continuous-time constraint violations.
             Added by propagate_trajectory_results.
+        parameters (dict[str, np.ndarray]): Parameter values the solve used, recorded
+            by :meth:`~openscvx.problem.Problem.solve` and
+            :meth:`~openscvx.problem.Problem.solve_batched` (batched solves store every
+            value with a leading ``(B,)`` axis, shared values replicated). Results
+            record what happened: post-processing propagates with these values, and
+            the Problem's parameter dict only seeds the next solve. Empty when no
+            snapshot was recorded (e.g. ``solve_jax``, whose callers manage
+            parameters themselves).
         plotting_data (dict[str, Any]): Flexible storage for plotting and application data.
 
     !!! note "JAX pytree registration"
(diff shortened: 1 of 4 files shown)


spadafab (approved):
Looks good

PR #548, opened 2026-07-06
Fix/batched-postprocess-tau-clamp

griffin-norris:
## Summary

`post_process_batched` clamped the **total** dense-grid size to `settings.prp.max_tau_len` — but `initialize()` sizes `max_tau_len` as the propagation solver's **per-segment** tau buffer width. Applying a per-segment capacity to the total point count over-clamps by roughly the number of segments: batched dense trajectories collapsed to ~80 samples where the sequential `post_process()` path produces ~1500 for the same problem.

This replaces the total-count clamp with a per-segment-aware bound. A uniform `linspace` grid drops about `(n_times − 1) · seg_dur / span` points into each segment, plus the endpoint sample `simulate_nonlinear_time` appends, so the bound solves

```
(n_times − 1) · frac_max + 2 ≤ max_tau_len
```

where `frac_max` is the largest segment-duration fraction across every segment of every batch element (from each element's node times via `s_to_t`). Near-uniform batches (`frac_max ≈ 1/(N−1)`) now pass through effectively unclamped; genuinely tight per-segment cases are still reduced enough that no segment can overfill the compiled buffer (verified empirically: a 100k-point request clamps to a grid whose busiest segment fills the buffer with one slot to spare).

## Context

Found while converting the race-car hybrid example (on `Chore/comparison` / `feature/f1-2026-energy`) to `solve_batched` + `post_process_batched` (built on #547): the three-car batch came out with 77-sample trajectories vs ~1500 sequentially, making the Viser replay visibly coarse. No caller-side workaround exists — `max_tau_len` is also the compiled pad width, so overriding it crashes with a shape mismatch.

## Tests

New `tests/test_post_process_batched_tau_clamp.py`:
- `test_batched_dense_grid_exceeds_per_segment_capacity` — default grid exceeds `max_tau_len` in total while every segment stays within capacity (fails before the fix: grid pinned to exactly `max_tau_len`).
- `test_batched_large_n_times_request_is_clamped_to_per_segment_capacity` — an absurd `n_times=100_000` is clamped and no segment overfills.

Full batched suites pass (57 tests across `test_post_process_batched_tau_clamp`, `test_solve_batched_brachistochrone`, `test_results_parameters`, `test_solve_batched_inference`, `test_solve_batched_cvxpy_export_errors`, `test_solve_batched_export_roundtrip`), plus `ruff check` / `ruff format` clean.

https://claude.ai/code/session_01RmUa9YhgoAidsUFm36DZTY

### DIFF
--- a/openscvx/problem.py
+++ b/openscvx/problem.py
@@ -63,7 +63,11 @@ from openscvx.lowered.jax_constraints import (
     LoweredJaxConstraints,
     LoweredNodalConstraint,
 )
-from openscvx.propagation import get_propagation_solver, propagate_trajectory_results
+from openscvx.propagation import (
+    get_propagation_solver,
+    propagate_trajectory_results,
+    s_to_t,
+)
 from openscvx.solvers import ConvexSolver, resolve_solver_config
 from openscvx.symbolic.builder import preprocess_symbolic_problem
 from openscvx.symbolic.expr import CTCS, Constraint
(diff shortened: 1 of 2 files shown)


haynec (approved):
LGTM

PR #549, opened 2026-07-06
Reorganize the test suite into unit / e2e / extras / examples groups

### LINKED ISSUE #541: 3rd Party Integration Tests
Moves `Frax`, `Mujoco`, `stljax`, `jaxlie`, and `cvxpygen` tests into standalone actions which are run on the same concurrency as `integration_tests`.

griffin-norris:
Closes #541

## What

Splits the test suite into four explicitly selectable groups and restructures CI around them:

| Group | Selection | Where it runs |
|---|---|---|
| Unit (~1,736) | `-m "not e2e and not examples and not extras"` | `tests.yml` `unit` job, 2 OS × py3.11–3.14 |
| End-to-end SCP solves (68) | `-m "e2e and not extras"` | `tests.yml` `e2e` job, gated on `unit` |
| Optional-dependency tests (142) | `-m "extras and not examples"` | `tests.yml` `extras` job, gated on `unit`, single cell with `.[test,qpax,mjx,lie]` |
| Examples sweep (61) | `-m examples` | `tests-examples.yml`, weekly / release / dispatch |

The four selections partition the suite exactly (1736 + 68 + 142 + 61 = 2007; verified disjoint and total).

## How

- **Marker-as-guard**: optional-dependency tests declare a plain per-package marker (`mjx`, `qpax`, `cvxpygen`, `lie`, `moreau` — names mirror the pyproject extras). A single conftest hook derives the umbrella `extras` marker and a skip-with-install-hint when the package is missing. This replaces every ad-hoc guard idiom (`importorskip`, `tests/_marks.py`, per-file skipifs), and backend params inside the solve tests use `pytest.param("qpax", marks=pytest.mark.qpax)` so they're individually selectable. Forgetting a marker now fails loudly in the unit job instead of silently skipping.
- **`tests/e2e/`**: the ten top-level full-solve files (brachistochrone/Hohmann family) move there and are auto-marked by the same hook; SCP-loop tests inside mirrored dirs (`tests/algorithms/`, `tests/solvers/`) stay put and carry `pytestmark = pytest.mark.e2e`.
- **`integration` → `examples`**, with discovery rewritten as a path-only glob: examples are imported inside their own test instead of at collection. Collection no longer constructs every example `Problem` (or mutates global JAX config), and a `sys.exit` in an example (e.g. `examples/frax/panda_frax.py` without frax) is one failed test instead of an aborted collection.
- **`test` extra slimmed** to `pytest`, `pytest-xdist`, `scipy`; `jaxlie`/`frax`/`svgpathtools` leave it (jaxlie-dependent tests carry the `lie` marker and run in the extras job).

## Notably fixed

- qpax had **zero CI coverage** (not installed in any workflow env); the extras job now runs the qpax solver tests and the qpax params of the e2e solves.
- Local `pytest --collect-only` no longer aborts with INTERNALERROR when frax is absent.

- **Per-integration badges**: a reusable `_extra.yml` workflow plus thin `tests-mjx.yml` / `tests-lie.yml` callers (push-to-main + weekly cron, not per-PR) feed badges in a new README **Third-Party Integrations** section, which lists every extra with its badge or tracking issue (#550, #551). Badges populate on the first post-merge run on main.

## Notes for review

- **qpax is omitted from the CI install lines** (#550): the extras job's first-ever qpax runs showed neither version is green on Linux runners — 0.1.3 breaks solve parity/convergence (reproduces locally too; note the `qpax` extra resolves to it for users), and 0.1.1 stalls convergence badly enough to blow the job timeout. Its tests skip with the install hint until #550 is resolved. Those first runs also caught `test_brachistochrone_multishot_roundtrip` bit-rotted against four API changes (fixed) and an uncalibrated 15s init budget (now 30s).
- **cvxpygen is omitted from the CI install lines** (#551): the `cvxpygen` extra is unsatisfiable while cvxpy is pinned `<1.9` (#507) — `cvxpygen>=1.0.0` needs `cvxpy>=1.9`. The old weekly integration job had been failing on this resolve since May; its tests now skip visibly with the install hint instead. Re-add the extra once the pins reconcile.
- Job display names keep the old `Unit Tests (Python X, os)` pattern so existing required status checks still match. If the gated `E2E Tests (...)` / `Extras Tests (Python 3.13, ubuntu-latest)` jobs should be required, that's a branch-protection settings change.
- moreau tests keep skipping in CI (license key); they're grouped under the `moreau` marker for when that changes.
- `stl` / `frax` markers are deliberately not registered yet — added when their first tests exist. The stljax bridge (`openscvx/symbolic/expr/stljax.py`) currently has no tests; follow-up candidate from #541.


haynec (changes requested):
LGTM do we want to have separate ci indicators in the readme for the different 3rd party integrations?

haynec (approved):
LGTM

PR #553, opened 2026-07-06
Memoize canonicalize and JAX lowering by node identity

spadafab:
### What changed (high level)

Symbolic expressions form a DAG: a reused intermediate (a DCM, a forward-kinematics transform, an aero force) is a single node with many parents. Both `canonicalize()` and the JAX lowerer used to walk this structure as a *tree*, re-processing every shared node once per path from the root — waste that compounds multiplicatively with nesting depth. This PR memoizes both traversals by node identity, so every node is processed exactly once regardless of how many parents share it. Same results, strictly less work; no user-facing API changes.

### How it works (low level)

Four pieces, in three files:

- **`canonicalize()` memo** (`openscvx/symbolic/expr/expr.py`): `Expr.__init_subclass__` wraps each subclass's `canonicalize()` with a per-thread, per-pass memo keyed on `id(self)`. A shared node is canonicalized once and the same object is returned to every parent, preserving sharing in the output. The top-level call owns the memo and clears it on exit; nodes are held alive in the memo so their `id` can't be recycled mid-pass, with an identity check on lookup.
- **Node cache** (`openscvx/symbolic/lowerers/jax/_lowerer.py`): `JaxLowerer.lower()` caches the lowered closure per AST node (`id(expr) → (expr, closure)`), so a shared node is lowered once instead of once per parent.
- **Per-trace value memo** (`_memoize_call`, same file): each lowered closure caches its last result on argument identity. Within a JAX trace every node receives the same argument objects, so a shared subexpression emits its jaxpr subgraph once instead of once per consumer. The cache only engages while tracing (an argument is a JAX tracer); eager calls and keyword-argument calls bypass it.
- **Trace-boundary guards** (`openscvx/symbolic/lowerers/jax/logic.py`, `openscvx/symbolic/lowerers/jax/vmap.py`): `lax.cond` branches and `vmap` bodies are traced as nested sub-traces while reusing the captured `(x, u, node, params)`. The visitors bracket these with `pause_memo()`/`resume_memo()` so a value cached in one trace is never returned inside another.

### Performance

Node counts (dynamics + constraints; "tree-expanded" is what the un-memoized traversal effectively processes, "DAG" is unique nodes):

| Example | Stage | Tree-expanded (before) | DAG (after) | Reduction |
|---|---|---|---|---|
| 7_dof_arm | raw | 1,192 | **113** | **10.5×** |
| 7_dof_arm | after canonicalize | 1,220 | 497 | 2.5× |
| 6DoF_pdg | raw | 1,600 | **340** | **4.7×** |
| 6DoF_pdg | after canonicalize | 1,617 | 441 | 3.7× |

Initialization time (import through `problem.initialize()`, two fresh-process runs each, CPU JAX):

| Example | Phase | After | Before | Speedup |
|---|---|---|---|---|
| 7_dof_arm | `initialize()` | **39.9 s** | 53.2–56.9 s | **~1.35×** |
| 6DoF_pdg | `initialize()` | **4.9–5.1 s** | 8.2–8.3 s | **~1.7×** |

Symbolic build time is unchanged (the fix affects traversals, not expression construction), and the remaining `initialize()` cost is dominated by XLA compilation.

### DIFF
--- a/openscvx/symbolic/expr/expr.py
+++ b/openscvx/symbolic/expr/expr.py
@@ -63,14 +63,50 @@ Example:
         canonical = expr.canonicalize()  # Simplifies to: x + x
 """
 
+import functools
 import hashlib
 import struct
+import threading
 from typing import TYPE_CHECKING, Callable, List, Tuple, Union
 
 if TYPE_CHECKING:
     from .linalg import Transpose
 
 import numpy as np
+
+# Per-thread memo for canonicalize(). The memo canonicalizes each node once and
+# returns the same object to every parent, preserving sharing.
+_canon_state = threading.local()
+
+
+def _memoized_canonicalize(fn: Callable) -> Callable:
+    """Wrap a subclass ``canonicalize`` so shared subexpressions are reused.
+
+    Keyed by ``id(self)``; the node is held alive in the memo for the duration of
+    the pass so its id cannot be recycled, with an identity guard on lookup. The
+    top-level call owns the memo and clears it on exit, so each independent
+    canonicalize pass starts fresh.
+    """
+
+    @functools.wraps(fn)
+    def wrapper(self):
+        memo = getattr(_canon_state, "memo", None)
+        owns = memo is None
+        if owns:
+            memo = {}
+            _canon_state.memo = memo
+        try:
+            entry = memo.get(id(self))
+            if entry is not None and entry[0] is self:
+                return entry[1]
+            result = fn(self)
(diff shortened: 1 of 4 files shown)


griffin-norris:
@spadafab This is a fantastic find! I think this means that I should set up some proper benchmarking to check for these kinds of performance issues in the future. I will open an issue to track this further and take a look at the code later tonight

griffin-norris (approved):
LGTM assuming CI passes

PR #560, opened 2026-07-11
Per-constraint licq_max overrides, remove licq_min

### LINKED ISSUE #558: Add support for per-constraint LICQ limits

### LINKED ISSUE #559: Remove LICQ min limit

griffin-norris:
Closes #558, closes #559.

## What changed

**Per-constraint LICQ limits (#558).** `CTCS`, the `ctcs()` helper, and every `.over()` method that seals to CTCS (`Constraint`, the STL exprs in `stl.py`, and `stljax`) now accept a `licq_max` bounding the augmented state that accumulates the constraint's violation penalty.

The bound for each augmented state is resolved as:
1. An explicit `licq_max` on a member constraint — all explicit values within a group must agree, otherwise a `ValueError` suggests aligning the values or splitting the group via `idx`.
2. Otherwise the problem-wide `Problem(licq_max=...)` default.
3. Which itself defaults to `1e-4`.

The dict-keyed-by-idx form of the `Problem` kwarg is gone — per-constraint `licq_max` is now the one way to set per-group bounds. The value is preserved through `canonicalize()` / `over()`, included in expression hashing, validated positive, and shown in `repr()`. The byof path is untouched — its specs already carry their own per-group `bounds`.

**Remove `licq_min` (#559).** Augmented states are now always bounded below at 0, with the initial condition and trajectory guess pinned there too.

## Notes

- Examples keep their original `Problem(licq_max=...)` style — that kwarg still means what it meant, so the only example change is dropping `licq_min=0.0` from `relative_loitering.py`.
- Auto-added time-bound CTCS constraints carry no explicit value, so they inherit the group's explicit value when sharing an augmented state with user constraints, and the problem-wide default otherwise — matching prior behavior.

## Verification

- `tests/symbolic/` (1529 tests) and `tests/e2e/test_brachistochrone.py` (39 tests) pass, plus ruff.
- New unit tests: per-group values, problem-default + per-constraint override precedence, explicit-wins-within-group, conflicting-values error, non-positive rejection (per-constraint and problem-wide).
- Dubins car, waypoint-STL (problem-wide `1e-10`), and flappy bird examples run to convergence.

### DIFF
--- a/examples/spacecraft/relative_loitering.py
+++ b/examples/spacecraft/relative_loitering.py
@@ -353,7 +353,6 @@ def build_relative_loitering_problem(
         time=time,
         constraints=constraints,
         N=n_nodes,
-        licq_min=0.0,
         licq_max=1e-8,
         discretizer={
             "ode_solver": "Dopri8",
--- a/openscvx/problem.py
+++ b/openscvx/problem.py
@@ -340,8 +340,7 @@ class Problem:
         dynamics_prop: Optional[dict] = None,
         states_prop: Optional[List[State]] = None,
         algebraic_prop: Optional[dict] = None,
-        licq_min: Union[float, Dict[int, float]] = 0.0,
-        licq_max: Union[float, Dict[int, float]] = 1e-4,
+        licq_max: float = 1e-4,
         algorithm: Optional[Union[Algorithm, dict]] = None,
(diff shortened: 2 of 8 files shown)


haynec (approved):
LGTM

PR #571, opened 2026-07-17
Fix/initialize sync guesses

### LINKED ISSUE #561: initialize() silently ignores State/Control .guess set after Problem construction
## Summary

Trajectory guesses are baked into the lowered problem when `Problem(...)` is constructed. Mutating `state.guess` / `control.guess` afterwards and then calling `initialize()` + `solve()` / `solve_batched()` silently uses the *old* guesses — the override is ignored with no warning. Only `reset()` re-reads them (via `_sync_guesses()`, `problem.py:853`).

## Reproduction

```python
D_throt.guess = 0.9 * np.ones((N, 1))
problem = ox.Problem(..., states=[..., D_throt, ...], ...)

D_throt.guess = trail_brake_profile          # takes no effect...
problem.initialize()
results = problem.solve_batched(parameters=cars)  # ...solves from the 0.9 guess
```

Discovered while A/B-testing initial guesses on `examples/race_cars/race_car_hybrid.py`: two runs with "different" guesses produced bit-identical lap times across five continuation rounds, because the post-construction override never reached the solver. Inserting `problem.reset()` after `initialize()` is the workaround:

```python
D_throt.guess = trail_brake_profile
problem.initialize()
problem.reset()      # _sync_guesses() — now the new guess is used
```

## Why this is surprising

- `initialize()` is documented as "prepare for optimization", and it *does* freshly build the algorithm state — just from stale unified guesses. Nothing in its docstring says guesses were snapshotted at construction time.
- `solve()` re-syncs parameters and boundary conditions on every call (`_sync_parameters()`, `_sync_boundary_conditions()`, `problem.py:1371-1373`), so `Parameter.value` and `State.initial/final` edits *do* take effect — guesses are the one input that doesn't follow the same mental model.
- The MPC warm-start pattern in the `reset()` docstring (`problem.py:1275`) shows `.guess` mutation is an intended workflow, which makes the `initialize()` path's silence more of a trap.

Per CLAUDE.md, hidden call-ordering requirements are an API bug: a user shouldn't need to read `_sync_guesses()` to learn that `reset()` is the only sync point.

## Proposed fix

Call `_sync_guesses()` (and arguably `_sync_boundary_conditions()`, for symmetry with `reset()`) inside `initialize()` before the algorithm state is created. Since `reset()` already does exactly this after `initialize()` has run, the change should be a one-liner plus a docstring note.

Open question: should `solve()` also sync guesses alongside its existing `_sync_parameters()` / `_sync_boundary_conditions()` calls? That would make all three user-editable inputs (parameters, boundary conditions, guesses) behave uniformly, but it changes behavior for anyone relying on `solve()` reusing the previous algorithm state mid-continuation, so it may deserve separate consideration.

griffin-norris:
Closes #561.

## Problem

Trajectory guesses were baked into the lowered problem at `Problem(...)` construction. Mutating `state.guess` / `control.guess` afterwards and calling `initialize()` + `solve()` silently used the *old* guesses — only `reset()` re-read them via `_sync_guesses()`. This broke the natural A/B mental model, where `Parameter.value` and `State.initial`/`final` edits *do* take effect on the next solve.

## Fix

`initialize()` now calls `_sync_guesses()` and `_sync_boundary_conditions()` before building the fresh state pytree — exactly mirroring `reset()`. Root cause: `settings.sim.x`/`.u` *are* `lowered.x_unified`/`u_unified`, and `AlgorithmState.from_settings` seeds the state from `settings.sim.x.guess`/`u.guess`, but nothing copied a post-construction `.guess` edit into the unified rep except `_sync_guesses()`.

The issue's open question — should `solve()` also sync guesses? — is resolved **no**: `solve()` never rebuilds the state pytree (it continues from the existing state for warm-starting/continuation), so syncing there would be a silent no-op, and forcing a rebuild would discard the warm-start. `initialize()`/`reset()` are the correct sync points since both create a fresh state.

## Tests

`tests/test_initialize_syncs_guesses.py` — a guess/boundary edited after construction reaches the solver state, and `initialize()` agrees with the historical `reset()` workaround. Verified the two `initialize` guards fail on baseline and pass with the fix.

### DIFF
--- a/openscvx/problem.py
+++ b/openscvx/problem.py
@@ -1113,6 +1113,8 @@ class Problem:
 
         This method vmaps dynamics, JIT-compiles constraints, builds the convex
         subproblem, and initializes the solver state. Must be called before solve().
+        The solver state is seeded from the current `.guess` and `.initial`/`.final`
+        on the State/Control objects; to re-seed after `initialize()`, use :meth:`reset`.
 
         Example:
             Prior to calling the `.solve()` method it is necessary to initialize the problem
@@ -1331,6 +1333,11 @@ class Problem:
         else:
             # Printing was disabled after __init__, disable emitter to avoid queue buildup
             self.emitter_function = lambda data: None
+
+        # Re-read guesses and boundary conditions off the State/Control objects
+        # before seeding the state, so post-construction edits take effect (as reset() does).
+        self._sync_guesses()
+        self._sync_boundary_conditions()
 
         # Create fresh solver state + history pair.
         self._state = self._default_state()
--- a/tests/test_initialize_syncs_guesses.py
+++ b/tests/test_initialize_syncs_guesses.py
@@ -1,0 +1,71 @@
+"""``Problem.initialize`` seeds the solver state from current guesses/boundaries.
+
+Trajectory guesses and boundary conditions live on the ``State``/``Control``
+objects the user builds. They are only copied into the lowered representation by
+``_sync_guesses`` / ``_sync_boundary_conditions``. Before #561, those syncs ran
+only inside ``reset()``, so a ``.guess`` edited between ``Problem(...)`` and
+``initialize()`` was silently ignored — the fresh state was seeded from the
+guess snapshotted at construction. ``initialize()`` now re-reads both, matching
+the ``.value`` / ``.initial`` mental model where post-construction edits take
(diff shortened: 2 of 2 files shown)


haynec (approved):
LGTM

PR #572, opened 2026-07-18
Corrected and verified mistaken FOH parameterization

### LINKED ISSUE #584: FOH blend weight reaches N/(N-1) instead of 1: reported nodal controls are not the applied control
## Summary

The FOH blend weight used to interpolate controls inside a segment reaches `N/(N-1)` at the segment end instead of `1`. The control that is actually integrated therefore overshoots the next node's value and jumps discontinuously back at each node, so **the nodal control values a user reads out of `results` are not the control the solver applied**. The discretizer and the propagator share the convention, so the solver remains internally self-consistent — states and `t_final` are correct — but `results.nodes[...]`, `results.u_full`, and `results.trajectory[...]` do not mean what first-order hold says they mean. The discrepancy scales as `1/(N-1)` and is worst at small `N`.

## Mechanism

The blend weight is computed as

```python
beta = tau * N * foh_mask
u = u_cur + beta * (u_next - u_cur)
```

at these sites:

- `openscvx/discretization/discretize_linearize.py:132`
- `openscvx/discretization/discretize_linearize.py:363`
- `openscvx/discretization/linearize_discretize.py:240`
- `openscvx/discretization/linearize_discretize_sparse.py:185`
- `openscvx/propagation/propagation.py:64` (as `(tau - tau_init) * N * foh_mask`)

but each segment is integrated over `tau` in `[0, 1.0/(N - 1)]` — see the integration horizon passed at `openscvx/discretization/discretize_linearize.py:150` and `:159`:

```python
sol = solve_ivp_rk45(multiple_dxdt, 1.0 / (N - 1), x.flatten(), ...)
```

At the end of a segment `tau = 1/(N-1)`, so `beta = N/(N-1)` rather than `1`. The control actually applied on segment `k` is

```
u(t) = u_k + (N/(N-1)) * ((t - t_k)/dt) * (u_{k+1} - u_k)
```

i.e. it overshoots `u_{k+1}` by `(u_{k+1} - u_k)/(N-1)` and then jumps back down to `u_{k+1}` at the node.

The tau grid is `np.linspace(0, 1, N)` (`openscvx/propagation/propagation.py:183`), which is consistent with segments of tau-length `1/(N-1)`, so the `1/(N-1)` horizon looks like the intended convention and the `* N` factor looks like the error.

## Numerical evidence

Using `examples/abstract/brachistochrone.py`, I solved at several `N`, then re-integrated each solve's own nodal controls through the example's dynamics with `scipy.integrate.solve_ivp` (rtol/atol 1e-12) two ways, and compared the endpoint against the solver's own `X[-1]`:

| N | error, textbook FOH | error, `beta = tau*N` model |
|---|---|---|
| 2 | 9.245e+00 | 9.98e-05 |
| 3 | 3.668e+00 | 1.55e-04 |
| 5 | 9.459e-01 | 1.37e-05 |
| 10 | 1.815e-01 | 6.54e-06 |
| 20 | 4.003e-02 | 2.52e-06 |

The `beta = tau*N` model reproduces the solver's own converged states to <= 1.6e-04 at every `N`, while textbook FOH is off by up to 9.2. This confirms the overshoot is what is being integrated, and that it is not merely a reporting artifact downstream of the solve.

### `n = 2` case in detail

The example ships with `n = 2`. There:

- `results.nodes["theta"]` / `results.u_full` / `results.trajectory["theta"]` report a ramp **0.895 deg -> 50.436 deg**.
- The control actually integrated ramps **0.895 deg -> 99.977 deg** (exactly `theta_0 + 2 * delta` since `N/(N-1) = 2`).
- Integrating the *reported* control honestly lands at `(8.045, -2.145)` instead of the imposed terminal condition `(10, 5)`.

For reference, the analytic brachistochrone through these boundary conditions has `theta(t) = 0.973846 * t` with `theta_f = 100.507 deg` and `t_f = 1.801295` — the *applied* control matches this; the *reported* control has half the slope.

### Energy-invariant check

`v^2/2 = g*(y0 - y)` holds along any trajectory of these dynamics regardless of the control, so it is a control-free consistency check on the post-processed output. Max `|v^2/2 - g*(10 - y)|` on `results.trajectory`:

- `N = 2`: **1.9e-02**
- `N = 20`: **7.8e-11**

The post-processed dense trajectory at `N = 2` is not an exact integration of the dynamics for the control it reports alongside it.

### Likely related symptom

At `N = 3` on this example the applied `theta` peaks at **106.96 deg** against a declared bound of `theta.max = 100.5 deg` — a 6.5 deg excursion past the control bound, with CTCS violation 1.08e-02. Because the bound is imposed at the nodes and the overshoot happens strictly between them, the per-segment overshoot appears to be able to carry the applied control outside its own declared box. This is the largest overshoot in the `N` sweep, consistent with `N/(N-1) = 1.5` at `N = 3`.

## Reproduction

Save as `repro.py` in the repository root and run `python -u repro.py`:

```python
import numpy as np
from scipy.integrate import solve_ivp

from examples.abstract.brachistochrone import problem  # n = 2

problem.initialize()
res = problem.solve()

U = np.asarray(res.U[-1])   # nodal controls, shape (N, n_u); column 0 is theta
X = np.asarray(res.X[-1])   # nodal states
tf = float(np.asarray(res.t_final).reshape(-1)[0])
N = U.shape[0]
tn = np.linspace(0.0, tf, N)
g = 9.81


def rhs(t, s, theta):
    x, y, v = s
    a = theta(t)
    return [v * np.sin(a), -v * np.cos(a), g * np.cos(a)]


def foh(t):
    """Textbook FOH: linear between nodes, hits u_{k+1} at t_{k+1}."""
    return np.interp(t, tn, U[:, 0])


def applied(t):
    """What OpenSCvx integrates: beta = tau * N over segments of tau-length 1/(N-1)."""
    k = int(np.clip(np.searchsorted(tn, t, side="right") - 1, 0, N - 2))
    beta = (t - tn[k]) / (tf / (N - 1)) * (N / (N - 1))
    return U[k, 0] + beta * (U[k + 1, 0] - U[k, 0])


print(f"N = {N}   t_f = {tf:.9f}")
print(f"reported nodal theta (deg): {np.degrees(U[:, 0])}")
print(f"applied theta at t_f (deg): {np.degrees(applied(tf * (1 - 1e-12))):.4f}")
print(f"solver X[-1][:3]          : {X[-1, :3]}")
for name, th in (("FOH        ", foh), ("beta=tau*N ", applied)):
    sol = solve_ivp(rhs, [0, tf], [0.0, 10.0, 0.0], args=(th,), rtol=1e-12, atol=1e-12,
                    max_step=tf / (N - 1) / 50)
    err = np.linalg.norm(sol.y[:3, -1] - X[-1, :3])
    print(f"  re-integrated with {name}: endpoint = {sol.y[:3, -1]}   |err vs X[-1]| = {err:.3e}")

# Energy invariant v^2/2 = g*(y0 - y) holds for ANY control.
res = problem.post_process()
pos = np.asarray(res.trajectory["position"])
vel = np.asarray(res.trajectory["velocity"]).reshape(-1)
print(f"  max |v^2/2 - g*(10 - y)| on post-processed trajectory: "
      f"{np.abs(vel**2 / 2 - g * (10 - pos[:, 1])).max():.3e}")
```

Observed output:

```
N = 2   t_f = 1.801348729
reported nodal theta (deg): [ 0.89451868 50.4355823 ]
applied theta at t_f (deg): 99.9766
solver X[-1][:3]          : [10.          5.          9.90462591]
  re-integrated with FOH        : endpoint = [ 8.04512014 -2.14465192 15.43625831]   |err vs X[-1]| = 9.245e+00
  re-integrated with beta=tau*N : endpoint = [9.99992814 4.99993222 9.90461155]   |err vs X[-1]| = 9.982e-05
  max |v^2/2 - g*(10 - y)| on post-processed trajectory: 1.907e-02
```

Environment: run on branch `feature/with-jacobian` (HEAD `3b441b6d`, version `0.5.1.dev1`); all five affected sites verified byte-identical on `main`. JAX 0.10.0 CPU backend, `float_dtype="float64"`, `VectorizeDiscretizeLinearize` (Tsit5), QOCO via `CVXPyPTRSolver`.

## Suspected fix direction

`beta = tau * (N - 1)` would make the blend weight reach exactly `1` at the end of a segment of tau-length `1/(N-1)`, giving a continuous piecewise-linear control that interpolates the nodes. I have not made this change or tested it.

Two things worth weighing before acting on that:

- The discretizer and the propagator would have to change **together**. They currently agree with each other, which is why converged solutions look self-consistent; changing only one side would make `post_process` disagree with the solve.
- Because the applied control changes, converged solutions will shift and tuned examples/tests may move — most visibly at small `N`. Worth checking whether any `expected` values in the test suite were fit to the current behavior.

No code changes are proposed here; filing for triage.

spadafab:
## Fix off-by-one in FOH interpolation weight (`tau*N` → `tau*(N-1)`)
Segments are integrated over `tau` in `[0, 1/(N-1)]`, but the FOH control interpolation weight was `beta = tau*N`, overshooting `u_next` by `N/(N-1)` at the segment end. The SCP still converges tightly (`J_vc ~ 1e-9`), but to trajectories consistent with a control ramp that's too steep by `1/(N-1)` — so the flown trajectory can deviate meaningfully from the certified one.

### Drone obstacle-avoidance example (N=6, tight tolerances, `licq_max=1e-6`)

Independently re-propagating the reported control law (correct FOH/ZOH convention, accurate integration) vs. what the solver certifies:

  | | before (bug) | after (fix) |
  |---|---:|---:|
  | converged | True (21 iters) | True (22 iters) |
  | true max obstacle violation | 0.136 | 0.010 |
  | final position error | 0.256 m | 0.000014 m |

The buggy build's certified trajectory looks collision-free at every node, but the *actually flown* path clips an obstacle; the fixed build clears it.

trajectory_zoom_xz

Fixes #584


haynec (dismissed):
LGTM, just need to resolve merge conflicts

griffin-norris (approved):
LGTM

PR #576, opened 2026-07-20
added qoco arg

haynec:
Changelog:
- Adds `"warm_start": false` to the default solver args, bypassing a known bug created when using `cvxpy>=1.9` and `qoco>=0.3.1`


griffin-norris (approved):
LGTM

PR #578, opened 2026-07-21
Add coefficient-override to Cinterp for between-solve retuning

griffin-norris:
## Why

`Cinterp` fits spline coefficients once at construction and bakes them into the compiled function as constants, so the only way to change an interpolated function is to redefine the problem and recompile. That is wasteful whenever a reference path or tabulated curve needs retuning between solves — warm-started MPC, parameter sweeps, and interactive examples where the user edits geometry while the SCP loop keeps running.

## What

Extends `Cinterp` itself (no parallel node) with a **coefficient override**: pass the `(4, n_seg)` polynomial table directly instead of `fp`, either as a numeric array or as a symbolic Expr — in practice an `ox.Parameter`:

```python
cs = PchipInterpolator(xp, fp_data)                     # any scipy method, host-side
coeffs = ox.Parameter("spline_coeffs", shape=(4, len(xp) - 1), value=cs.c[::-1])
ref = ox.Cinterp(progress[0], xp, coeffs=coeffs)

# ... between solves:
problem.parameters["spline_coeffs"] = PchipInterpolator(xp, new_fp).c[::-1]
```

A symbolic override is lowered like any Parameter — gathered from the params dict at run time — so swapping the table never recompiles. Fitting moves host-side, which is what makes the feature method-agnostic: cubic, PCHIP, Akima, or anything else that produces a piecewise-cubic table.

Design points:

- Exactly one of `fp` / `coeffs` is required; `method` alongside `coeffs` is rejected (it only governs fitting). `method`'s default became a `None` sentinel (normalized to `"cubic"` on the fit path) so an explicit `method` is distinguishable from a defaulted one.
- `xp` stays numeric/fixed — it is structural (drives the segment lookup).
- Symbolic coeffs join `children()` so parser traversal binds the Parameter; canonicalize, hashing, `repr`, and the LaTeX lowerer handle the new form. The numeric fit path is unchanged, including its hash bytes.
- SCP linearizes w.r.t. states and controls only, so no gradients ever flow through the fitting — only exact derivatives w.r.t. the query point, which come from the coefficients regardless of who computed them.
- With an override, continuity of the table becomes the caller's responsibility; the docstring warns that a discontinuous table degrades the discretizer silently.

## Tests

9 new tests in `tests/symbolic/expr/test_math.py`: scipy parity for cubic/PCHIP/Akima through a `Parameter`, parameter update between two evaluations of the same lowered function (no recompile), numeric-override regression, and validation errors (shape, both/neither, method+coeffs). Targeted runs: 29 passed (cinterp), 412 passed (math/latex/hashing/parameters), 67 passed (parser/jax lowering).

https://claude.ai/code/session_01AdMsrPCbaCSD4sHMkQta7a

### DIFF
--- a/openscvx/symbolic/expr/math.py
+++ b/openscvx/symbolic/expr/math.py
@@ -1135,23 +1135,40 @@ _CINTERP_METHODS = {
 
 
 class Cinterp(Expr):
-    """1D cubic spline interpolation for symbolic expressions.
+    """1D piecewise-cubic interpolation for symbolic expressions.
 
-    Computes a cubic spline through data points (xp, fp) and evaluates it at
-    query point x.  Spline coefficients are precomputed at construction time
-    via scipy so that JAX evaluation reduces to a segment lookup and
-    polynomial eval.
+    Evaluates a piecewise-cubic polynomial at query point ``x`` by a segment
+    lookup on the fixed breakpoints ``xp`` followed by Horner evaluation.  The
+    polynomial table is held as ``coeffs`` of shape ``(4, n_segments)`` where
+    ``n_segments = len(xp) - 1``; row ``k`` holds the degree-``k`` coefficients,
+    so the value in segment ``i`` at local coordinate ``t = x - xp[i]`` is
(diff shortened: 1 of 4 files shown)


haynec (approved):
LGTM

PR #581, opened 2026-08-06
Allow callable CTCS penalties built from symbolic expressions

griffin-norris:
Closes #420. First of a three-PR stack for the symbolic-layer ancillary-function family (#476); followed by the symbolic-guesses and with-jacobian PRs.

## What

`penalty` on `ox.ctcs` / `CTCS` / `constraint.over()` (and the STL `.over()` variants) now also accepts a callable `(residual: Expr) -> Expr`, applied once at problem-build time to the constraint's canonical residual:

```python
ox.ctcs(altitude >= 10, penalty=lambda r: ox.Huber(ox.PositivePart(r), delta=0.5))
```

The composed expression rides augmentation, autodiff, sparsity, and hashing exactly like the built-in penalties. This also makes `Huber`'s `delta` and `SmoothReLU`'s `c` tunable, which the string form never allowed. The string presets are unchanged and remain the default.

## Notes

- CTCS hashing now hashes the **built penalty expression** rather than the penalty's spelling, so a callable that rebuilds `Square(PositivePart(r))` hashes identically to `penalty="squared_relu"`. String-penalty hash values change (compile-cache invalidation only).
- Penalty building blocks (`PositivePart`, `Square`, `Huber`, `SmoothReLU`) are now exported at the `ox.` top level.
- Fixes a latent test bug: `test_hashing.py` used invalid penalty names (`"l2"`, `"l1"`) that only passed because hashing never built the AST.
- A callable returning a non-Expr (e.g. `jnp.maximum(r, 0)`) raises a build-time teaching error.
- Deliberately untouched: the byof penalty registry (`square`/`l1`/`huber`, divergent Huber default and formula) and the YAML string dialect.

## Testing

`tests/symbolic/` + expert/loader sweeps green (1716 at time of authoring); new tests cover the callable AST composition, parameter tuning, non-Expr rejection, repr, `.over()` round-trip, numeric equivalence against hand-built expressions, and hash equivalence of string vs. callable spellings. End-to-end brachistochrone with two callable penalties converges.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01M5WX9YVDJr8qmZd8z3DM1V

Issue #476, opened 2026-05-13
Additional Expr Usecases


griffin-norris:
- Add more usecases for the symbolic expression system
- May require modifications and generalizations to the lowering system to account different evaluation shapes

PR #582, opened 2026-08-06
Accept symbolic expressions and callables of tau as initial guesses

### LINKED ISSUE #460: Explore Passing Functions for Initial Guesses
- Could be useful to support symbolic functions that create the initial guess
- Provide the current initialization helpers as options

### LINKED ISSUE #475: Symbolic Expression Based Initialization
- Allow `Variable` initial guesses to be an `Expr` object

griffin-norris:
Closes #475 (and closes #460). Second in the #476 stack — **based on #581**; review that first.

## What

`Variable.guess` now takes, besides the existing `(N, n)` array:

```python
speed.guess = ox.linalg.Norm(vel)                  # Expr in other states/controls
pos.guess   = lambda tau: p0 + (pf - p0) * tau     # callable of the normalized node coordinate
```

A symbolic guess defines the guess **per node**: it is evaluated at every node with each referenced variable at its own guess value and tau running 0 → 1 — the same mental model as dynamics and constraints (so `ox.Norm(vel)` is the per-node speed of the velocity guess). The expression is the dependency graph: referenced guesses resolve first, cycles and foreign variables raise teaching errors. Re-assigning a symbolic guess between solves works through `reset()`/`initialize()` (MPC pattern).

## Design notes

- Evaluation reuses the unmodified lowering pipeline: `jax.vmap(lower_to_jax(expr), in_axes=(0,0,0,None))` over the packed guess trajectories, with the tau grid under a reserved `params` key. No visitor changed; one `Tau` leaf visitor was added.
- `Tau` is internal plumbing — the guess callable receives an instance; users never import it. Its `check_shape()` raises so a captured tau can't leak into dynamics/constraints.
- Resolution runs in the builder immediately after slice assignment and before augmentation derives the time-dilation guess from `time.guess`.
- Deliberately **no `Gradient` op**: trajectory-level constructions don't fit per-node evaluation and stay eager numpy (`vel.guess = np.gradient(pos.guess, axis=0)`), documented.
- `time.guess` accepts the symbolic forms; `time_dilation_guess` stays array-only. Propagation-only states reject symbolic guesses with a clear error.

## Testing

26 new tests (`tests/symbolic/expr/test_guess_expr.py`): setter dispatch, Tau mechanics, dependency ordering/cycles, per-node reductions, Parameter refs, shape coercion and errors, builder integration incl. time-dilation derivation, and Problem-level initialize/re-assign/reset. Symbolic sweep green.

https://claude.ai/code/session_01M5WX9YVDJr8qmZd8z3DM1V

### DIFF
--- a/docs/UsersGuide/01_hello_world_brachistochrone.md
+++ b/docs/UsersGuide/01_hello_world_brachistochrone.md
@@ -164,6 +164,35 @@ Here, we do _not_ specify an initial or final value for the control but we _do_ 
 The initial guess must be of shape $(n_u \times n_{\mathrm{nodes}})$, providing an initial guess for each control at every node.
 Technically, we can also provide a `.guess` for states. However, these default to a linear interpolation between initial and final conditions which is sufficient in most cases.
 
+#### Symbolic initial guesses
+
+`.guess` accepts three forms, on both states and controls:
+
+- an array of shape $(n_{\mathrm{nodes}} \times n)$, as above;
+- a **symbolic expression** in other states and controls;
+- a **callable of `tau`**, the normalized node coordinate, returning a symbolic expression.
+
+A guess expression is evaluated at every node, with each referenced variable at its guess value and `tau` running $0 \to 1$ across the grid — the same per-node mental model as dynamics and constraints.
+
+```python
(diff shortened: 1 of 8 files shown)


haynec (approved):
LGTM

